PR #73, opened 2021-12-31
README Refactor

ElectronicsArchiver:


ElectronicsArchiver:
@ethanaobrien bump

ethanaobrien:
Wow you're better at writing readme's than I am. Before I merge I have a few requests of things to change

1. Could you remove the following lines from the readme. There isn't much point because I now support every system on the site.

```
If your system is not listed here, navigate to EmulatorJS.com ,
check if it's available there and if it is, simply open an Issue .
```

2. is it possible that in the `supported systems` chart that it would have each system on one line instead of always having it wrap?

ethanaobrien:
@ElectronicsArchiver Looks good, just I'm wondering if its possible to still have it in the chart but just not wrap the lines on the chart (for bigger screens)

ElectronicsArchiver:
I don't think it's possible to have wrapping table lines in github markdown,
if it were proper html / css you could use flex elements I guess.

As a compromise I use inline wrapping in this last revision.

ElectronicsArchiver:
Github markdown renders have a set maximum width anyways,
except if used with Github Pages that is.

ethanaobrien:
Ok, seems good

PR #147, opened 2022-02-20
Update file naming

allancoding:
I updated the file naming. I was trying to find the audio element too but it did not work.
I also added index.html the emulator button.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+github.allancoding.ga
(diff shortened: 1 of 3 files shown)


ethanaobrien:
Thanks didnt think about doing some of that stuff! By the way I managed to fix the save/load state on the beta cores, so dont worry about that!

PR #154, opened 2022-02-23
Fixing index.html and adding translate.html

allancoding:
Fixing index.html and adding translate.html


ethanaobrien:
I really like the new style of the `index.html`. Could we make sure the background color stays the same as before the file is chosen? Also, could we center the created game element?

In the `translate.html` file, I think its a really cool idea but based on some of the issues I've gotten it might not be very user friendly. Do you think you could change it a bit to where it would ask you to translate one word at a time and you would either press enter of a next button to go to the next word?

allancoding:
What do you mean?

https://user-images.githubusercontent.com/74841470/155364808-f0705748-52e7-4df5-870b-4c6769545b7c.mp4

ethanaobrien:
Oh, I see what you mean. Thanks!

allancoding:
I am fixing the `index.html` too.

PR #173, opened 2022-03-10
Update index.html

MellonBunnu:
The background looks better lol

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,7 @@
     <body>
 	<style>
 body{
-  background: #c4bdff;
+  background: ##1ffffb;
 	margin: 0;
 	padding: 0;
 }


allancoding:
1. First there are two `##` in `background: ##1ffffb;`
2. Second there should be a 3rd opinion to change it.

PS: The background color is the same throughout the website. https://emulatorjs.ga/

ethanaobrien:
@allancoding I am pretty sure it is only 1 `#` and shouldnt we use the [`background-color` property](https://www.w3schools.com/cssref/pr_background-color.asp)?

ethanaobrien:
Either way, I really like the color it is currently set to. I like how it clashes with the logo. So sorry, but I'm closing this

You can always just host from your own site :)

PR #353, opened 2022-08-31
Auto Minify!

allancoding:
Auto Minify! sorry for all the commit's.


ethanaobrien:
This is actually going to be very useful, thanks! Two thing first, could you make another branch and commit the feature there to prevent the amount of commits? Also, I left a comment on one of the lines that might make it a bit cleaner

allancoding:
Ok will do.

allancoding:
I could not find the comment.

PR #359, opened 2022-09-02
Add new things

allancoding:
Add (ad) functions
And some other things


ethanaobrien:
What kind of things did you add?

allancoding:
The new `EJS_autoCloseAd` config if a number is set it will auto remove the ad in milliseconds if set to true removes right before game loads. and is false by default. It will override `EJS_noAutoCloseAd` 
added .ejs--backgroundimg for a css background image like so:
```
.ejs--backgroundimg{
    background-image: url("image-url") !important;
    background-color: transparent !important;
    background-size:contain;
    background-position: center;
    background-repeat: no-repeat;
}
```

ethanaobrien:
Can we have this as  `EJS_background`?

allancoding:
Yes we could That would be great but I was having a hard time trying to do that because that was my original idea. I don't if you could do that you might know how to do that better.

ethanaobrien:
I'll look into this when I can

Cattn (approved):

ethanaobrien:
@allancoding if you could, could you change the no close ad variable to an number (instead of having both) and having that the amount of milliseconds? If the input is below 1, then it will not auto close the ad. I'll look at the background feature when I can, sorry for the delay. I recently got a job so I haven't had much time for any side projects

PR #360, opened 2022-09-03
Update chi-CHI.json

eric183:
Optimize some Chinese translations in localization.json and fix some Chinese translation errors


eric183:
And maybe rename chi-chi.json to Zh-CN.json to make it readable, I think

ethanaobrien:
Thank you for the work! I'll rename the file with the next big update

Issue #477, opened 2023-01-31
"Insert before" Error. (Possible Fix included?)


Cattn:
```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'insertBefore')
    at loader.js:48:34
    at new Promise (<anonymous>)
    at loadScript (loader.js:36:16)
    at loader.js:60:15
(anonymous) @ loader.js:48
loadScript @ loader.js:36
(anonymous) @ loader.js:60
await in (anonymous) (async)
(anonymous) @ loader.js:97
```

I've been getting this error on some of my projects recently, and it doesn't seem to be anything I've done as even when removing all scripts other than the base EmulatorJS files I still get this.

At first I only got this error on 1 Project while using a Local Hosted Web Server, so I wrote it off as being my fault, but then it happened to me a few times on my live webpage. I do have a solution that fixed my issue, I'm making this issue in case it's a more widespread issue, and my solution may be able to help others.

My solution is to change the loadScript(file) function to include a check for the script.

```
function loadScript(file) {
    return new Promise(function (resolve, reject) {
        let scriptTag = document.getElementsByTagName('script')[0];
        if (!scriptTag) {
            console.error("Script tag not found, unable to load script");
            return;
        }
        let script = document.createElement('script');
        script.src = function() {
            if ('undefined' != typeof EJS_paths && typeof EJS_paths[file] == 'string') {
                return EJS_paths[file];
            } else if ('undefined' != typeof EJS_pathtodata) {
                if (!EJS_pathtodata.endsWith('/')) EJS_pathtodata+='/';
                return EJS_pathtodata+file+'?v='+VERSION;
            } else {
                return file+'?v='+VERSION;
            }
        }();
        scriptTag.parentNode.insertBefore(script, scriptTag);
        script.onload = resolve;
    });
}
```

for some reason this solved my issue, even though there was never no script tag
(The more I look at the solution the more questions I ask myself about, if this really solved my issue, or I just got lucky that it stopped? or it is maybe another, more niche issue? I'm not sure, it just doesn't feel like that would fix it)
#478 Pull RQ

PR #478, opened 2023-01-31
Insert Before Error Fix

Cattn:
Please refer to my post in #477 for more information on this.

TLDR: Not 100% sure if this is just a bandaid solution for a more niche problem, or this solves an issue preventing the emulator from loading sometimes


ethanaobrien:
Before I merge this, I'm really confused why this would be happening. How are there no script tags on the page? How are you loading the scripts? Are you manually deleting any tags? I'm very confused about how this is happening. What browser are you using?

Cattn:
> Before I merge this, I'm really confused why this would be happening. How are there no script tags on the page? How are you loading the scripts? Are you manually deleting any tags? I'm very confused about how this is happening. What browser are you using?

1. I'm really not sure how this happens, 
2. https://github.com/Math-Study/wii-gba/blob/main/launchpad.html happens here (it happened if I removed loadGame.js script, and just hard coded the link in normally as well.) it also happens to me on some of my other projects.
3. None of my scripts should be messing with tags, in the example above: https://github.com/Math-Study/wii-gba/blob/main/data/loadGame.js The JS file simply sets the values for game info.
4. I'm using Chrome V109 on Windows

I've had reports of this from some people using my sites, although it's been harder to track, and also seems to of popped up recently, I updated to the newest version of EmulatorJS a while ago so I don't think it's from that. I also don't think it's something I've done due to the reasons above

Cattn:
I've tried investigating why it happens, but I'm pretty confused, what I can say is that a hard reload of the page usually will fix it immediately while a normal one may take a few tries

Cattn:
To note* This solution I'm PR is implemented currently in the above linked repository

ethanaobrien:
Could you send me a link to a page that has the broken version? I would like to see this error.

Cattn:
> Could you send me a link to a page that has the broken version? I would like to see this error.

Yes, let me revert the changes one of my forks, and I'll send it, it's kinda like the freezing issue, it doesn't show up every time but it definitely is much more consistent

Cattn:
1. https://cattn.github.io/gba/gba.html I just noticed it seems to happen when coming from another page : Scroll down, then just click one of the games and it should work
2. https://math-study.github.io/wii-gba/ Just click on the New Super Mario Bros tile, and then play 

I've noticed you also might need to reload a few times before it happens.

My current theory after looking at all of this seems to be it may be because somehow of the values being set when going to the page?

This is seeming more like something that I've kinda caused, but also doesn't seem like it's very uncommon of an issue, or specific as both sites use completely different systems

ethanaobrien:
Would you mind updating the core files? I just pushed a change that should fix this. It turns out the insert before call wasnt even needed because the variables are still defined by the time we need them because of when the onload event fires.

Cattn:
> Would you mind updating the core files? I just pushed a change that should fix this. It turns out the insert before call wasnt even needed because the variables are still defined by the time we need them because of when the onload event fires.

Will do, one moment

Cattn:
I updated the Second Link

Cattn:
Sorry, I meant I updated the first link I listed above

Cattn:
Works without issue now, thanks!

ethanaobrien:
Awesome! I'll go ahead and close the linked issue

Cattn:
Gotcha

PR #607, opened 2023-08-10
rewind

n-at:
Adds rewind control. Disabled by default - not all cores may support rewind and it uses additional memory (may be problem on "heavy" platforms like psx).
Introduces new option `EJS_rewindEnabled`.

### DIFF
--- a/data/GameManager.js
+++ b/data/GameManager.js
@@ -67,8 +67,8 @@ class EJS_GameManager {
                "video_vsync = true\n" +
                "video_smooth = false\n" +
                "fastforward_ratio = 3.0\n" +
-                (window.EJS_rewindEnabled ? "rewind_enable = true\n" : "") +
-                (window.EJS_rewindEnabled ? "rewind_granularity = 6\n" : "") +
+                (this.EJS.rewindEnabled ? "rewind_enable = true\n" : "") +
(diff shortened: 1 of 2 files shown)


ethanaobrien:
Hey @n-at and thank you so much. I do have one question. Instead of having the site owner needing to enable Rewind, how about adding the option `Rewind (requires restart)` to the settings menu? Since Rewind (as you mentioned) may impact performance. 


Then just read this option from the saved settings and setup the emulator based on that.

n-at:
Yes, that's will be better. Added rewind control in settings menu.

ethanaobrien:
image
Is it safe to do this? Since rewind requires restart - shouldnt this variable only be set on load?

Other than that it looks good! Thank you so much!

n-at:
Yes, it's not needed. Removed that code.

ethanaobrien:
Awesome, thank you so much! I'll merge this right after I get the cores compiled, and thank you so much!

allancoding:
@n-at Thank you so much! In the future when adding a new option like  `EJS_rewindEnabled` add it to the loader.js
It will make adding documentation a lot easier. We didn't end up using that option, but i'm just telling you for future reference.

Thanks!

Issue #603, opened 2023-08-09
[Feature Request] Slowdown


deadman626:
Back when Online Emulators used to run on Adobe Flash Player I could easily use Cheat Engine to slow games down to easily dodge large amounts of enemy bullets or jump on fast-moving platforms or to win in Quick Time Events.

- Please add a Games Speed Sittings that lets me play games at Half or Quarter of game original Speed. it will add more fun to it.

ethanaobrien:
Oh wait, this isnt done yet

PR #608, opened 2023-08-11
slow motion

n-at:
Slow motion, control and settings.


n-at:
Will solve #603

ethanaobrien:
Thank you so much!

PR #595, opened 2023-08-07
atari2600, atari7800 - update controls and virtual gamepads

n-at:
For issue #591

atari2600
atari2600

atari7800
atari7800

PR #611, opened 2023-08-12
 game speed control buttons on predefined virtual controllers

n-at:
As discussed in https://github.com/EmulatorJS/EmulatorJS/pull/595

Example of nes controller with disabled rewind
wo-rewind

Example of nes controller with enabled rewind
with-rewind


ethanaobrien:
Thank you!

PR #612, opened 2023-08-13
Add new options and update pt-BR.json

allancoding:
New options added:
- `EJS_backgroundBlur`
- `EJS_backgroundColor`
- `EJS_AdMode`
- `EJS_AdSize`
- `EJS_adBlocked`

Options changed:
- `EJS_backgroundImage`
- `EJS_AdTimer`

`EJS_backgroundBlur` will override the `EJS_backgroundColor` the background color is the color of the background. The blur example is shown below:

You can set `EJS_backgroundBlur` to true and the default is not set.
You can set `EJS_backgroundColor` to a css color like `#000` or `rgba(150,150,150,0.6)`

`EJS_AdMode` has 3 possible values `0` `1` `2` 2 is the default
`0` will only show the ad on the start screen and not the loading screen
`1` will only show the ad on the loading screen
`2` will show on both

`EJS_adBlocked` is a function that can be run to change the adUrl on the fly or delete it.
Example: `EJS_adBlocked("https://ads.emulatorjs.org/blocked")`(to change the url) or `EJS_adBlocked("",true)`(to delete it)

`EJS_AdSize` is an array with the first value is the width and the second is the height. It can be any css option `%`, `vh`, `px` etc.
The ad is the top most thing so you can cover the start button by accident.
Example: `EJS_AdSize = ["800px", "500px"];`

`EJS_backgroundImage` was changed with the new options it now has to be a Absolute path or a Relative Path from the `pathtodata` folder

`EJS_AdTimer` is the same except if you set it to `-1` the ad will be remove when the game starts.

Fixed:
- Start Button animation
- Start Button border when using a image
- Loading text visibility when using image
- other css errors


ethanaobrien:
Why are we using an EventTarget to act upon the start button being clicked instead of the element itself? Or moving the code to the function that's being called either way? You could also add another event to emulatorjs, meaning how I have multiple `this.on` calls in the emulator scripts, id just rather not manually deal with events.

Could you also include screenshots of what each option does, and what the lack of the option does?

allancoding:
Are there any more screenshots that you would like to see?

allancoding:
And also added option:
- `EJS_ready`

For when the dom of EmulatorJS is ready
```js
EJS_ready = function() {
     console.log("EmulatorJS DOM is ready!");
     EJS_adBlocked("https://emulatorjs.org/?blocked");
}
```

PR #624, opened 2023-08-16
In-Game Auto Borders Script.

Diegrosan:
I liked the really cool project. I created a script to review, "bazels.js", an optional way to add automatic borders in the game, through thebezelproject, is really cool.
bezelsteste


allancoding:
This would be really cool but are some things that need to be cleaned up like removing the example image and comments, etc.

An idea is to modify the `EJS_backgroundImage` function and add `EJS_backgroundBezel`

The logic would be:
The `EJS_backgroundImage` is set use it on the loading screen and start screen.
If `EJS_backgroundImage` and `EJS_backgroundBezel` are set then the `EJS_backgroundImage` will only be on the start and loading screens. And `EJS_backgroundBezel` will show in game.
But if `EJS_backgroundBezel` is the only one set then `EJS_backgroundBlur` will be ignored and `EJS_backgroundBezel` will be shown on start, loading, and in game screens.
It would integrate the logic of the `bezels.js` into `emulator.js` file somewhere below the if statement shown below.

This is the if statement in the `emulator.js` file to help you get started: https://github.com/EmulatorJS/EmulatorJS/blob/main/data/emulator.js#L278-L288

Thanks

Diegrosan:
Okay, thank you very much.

PR #630, opened 2023-08-23
fix joystick and keyboard input localization

### LINKED ISSUE #626: Gamepad translations
How can I translate the labels inside Gamepad  configuration menu (pad directionals, stick directionals, and side buttons)?
I totally forgot this when I was checking the translation...

allancoding:
Should re fix #626


allancoding:
When I merge from now on I will squash the commits.

data/emulator.js
```diff
@@ -452,7 +452,10 @@ class EmulatorJS {
         this.elements.parent.appendChild(this.textElem);
     }
     localization(text, log) {
-        if (!isNaN(text)) return text;
+        if (!text) return text;
+        text = text.toString();
```
ethanaobrien, line 457: Why are we `toString`ing this?

data/emulator.js
```diff
@@ -452,7 +452,10 @@ class EmulatorJS {
...
+        if (text === " ") text = "space";
```
ethanaobrien: what does this do?

data/emulator.js
```diff
@@ -452,7 +452,10 @@ class EmulatorJS {
...
+        if (text === "") return text;
```
ethanaobrien: Will be caught by `!text`

data/emulator.js
```diff
@@ -2245,18 +2248,20 @@ class EmulatorJS {
                     textBox2.value = "";
                     textBox1.value = "";
                     if (this.controls[i][k] && this.controls[i][k].value !== undefined) {
-                        textBox2.value = this.controls[i][k].value;
+                        textBox2.value = this.localization(String(this.controls[i][k].value));
                     }
-                    if (this.controls[i][k] && this.controls[i][k].value2 !== undefined) {
-                        textBox1.value = this.controls[i][k].value2;
+                    if (this.controls[i][k] && this.controls[i][k].value2 !== undefined && this.controls[i][k].value2 !== "") {
+                        let value2 = String(this.controls[i][k].value2).split(":");
```
ethanaobrien: Can we call the tostring function instead of using the String constructor

data/localization/Translate.html
```diff
@@ -193,54 +203,148 @@ <h1>Translate Languages</h1>
             "R STICK UP": "-R STICK UP",
             "R STICK DOWN": "-R STICK DOWN",
             "R STICK LEFT": "-R STICK LEFT",
+<<<<<<< HEAD
```
ethanaobrien: merge conflicts

data/localization/af-FR.json
```diff
@@ -134,6 +144,23 @@
     "STICK DOWN": "COLLER",
     "STICK LEFT": "BÂTON GAUCHE",
     "STICK RIGHT": "COLLEZ À DROITE",
+<<<<<<< HEAD
```
ethanaobrien: Merge conflicts

data/localization/ar-AR.json
```diff
@@ -128,12 +138,33 @@
     "X": "X",
     "Y": "ص",
     "L": "إل",
+<<<<<<< HEAD
```
ethanaobrien: Im assuming every file downwards has merge conflicts

data/emulator.js
```diff
@@ -2573,188 +2578,188 @@ class EmulatorJS {
...
-                {"type":"button","text":"Select","id":"select","location":"center","left":-5,"fontSize":15,"block":true,"input_value":2}
+                {"type":"button","text":this.localization("Start"),"id":"start","location":"center","left":60,"fontSize":15,"block":true,"input_value":3},
+                {"type":"button","text":this.localization("Select"),"id":"select","location":"center","left":-5,"fontSize":15,"block":true,"input_value":2}
             ];
             info.push(...speedControlButtons);
         } else if ('nes' === this.getControlScheme()) {
             info = [
-                {"type":"button","text":"B","id":"b","location":"right","right":75,"top":70,"bold":true,"input_value":0},
-                {"type":"button","text":"A","id":"a","location":"right","right":5,"top":70,"bold":true,"input_value":8},
+                {"type":"button","text":this.localization("B"),"id":"b","location":"right","right":75,"top":70,"bold":true,"input_value":0},
```
ethanaobrien: This is a little messy, can we remove any calls to localization within these if statements and just cycle through the info object and set these afterwards

ethanaobrien (changes requested):

data/emulator.js
allancoding: To convert it to a string if it is a number

ethanaobrien: Then this should be before `if (!text)` since `if (!0)` will return true

data/emulator.js
allancoding: if it equals a space it changes the text to space, it is for the control mapping.

data/emulator.js
allancoding: Your right that can be removed.

data/emulator.js
ethanaobrien: but what does it do overall, how does it affect the outcome?

data/emulator.js
allancoding: Yes

data/emulator.js
allancoding: Yes lol that's a good idea, don't know why I did not think of that.

data/localization/Translate.html
allancoding: I will fix all of those

data/emulator.js
allancoding: Because if you map the space bar then it will show a space not the actual space text.

data/emulator.js
allancoding: I don't actually think it is necessary, I will just remove it.

data/emulator.js
ethanaobrien: Oh, can we change this in the keychange event then? We shouldnt do this here

data/emulator.js
ethanaobrien: Keep the check just in case

allancoding: Ok so it is good as is.

data/emulator.js
allancoding: Ok I'll try

data/emulator.js
ethanaobrien: (do `text = text.toString` first, and make sure `typeof text !== "undefined"`)

data/localization/ar-AR.json
allancoding: I think I fixed them all.

data/emulator.js
allancoding: lol I forgot this

allancoding:
That look good now?

data/emulator.js
```diff
@@ -452,16 +452,18 @@ class EmulatorJS {
...
     localization(text, log) {
-        if (!isNaN(text)) return text;
-        if (text.includes("EmulatorJS v")) return text;
-        if (this.config.langJson) {
-            if (typeof log === "undefined") log = true;
-            if (!this.config.langJson[text] && log) {
-                if(!window.EJS_missingLang) window.EJS_missingLang = [];
-                window.EJS_missingLang.push(text);
-                console.log("Translation not found for '"+text+"'. Language set to '"+this.config.language+"'");
+        if (typeof text !== "undefined"){
```
ethanaobrien: instead of making this a big if statement, can we reverse this check

```
if (typeof text === "undefined") return;
```

data/emulator.js
```diff
@@ -452,16 +452,18 @@ class EmulatorJS {
...
+            text = text.toString();
+            if (text.includes("EmulatorJS v")) return text;
+            if (this.config.langJson) {
+                if (typeof log === "undefined") log = true;
+                if (!this.config.langJson[text] && log) {
+                    if(!window.EJS_missingLang) window.EJS_missingLang = [];
```
ethanaobrien: (there should be a space after the if and before the `(`)

Can we define these within the emulatorjs instance?
`this.missingLanguages = []`. This should be initialized within the constructor

data/emulator.js
```diff
@@ -452,16 +452,18 @@ class EmulatorJS {
...
+                    window.EJS_missingLang.push(text);
```
ethanaobrien:
```
if (!EJS_missingLang.includes(text)) EJS_missingLang.push(text);
```

data/emulator.js
```diff
@@ -2245,18 +2247,24 @@ class EmulatorJS {
...
+                    if (this.controls[i][k] && this.controls[i][k].value2 !== undefined && this.controls[i][k].value2 !== "") {
+                        let value2 = this.controls[i][k].value2.toString().split(":");
+                        textBox1.value = this.localization(value2[0]) + ":" + this.localization(value2[1]);
                     }
                 })
                 
                 if (this.controls[i][k] && this.controls[i][k].value) {
-                    textBox2.value = this.controls[i][k].value;
+                    let value = this.controls[i][k].value.toString();
+                        if (value === " ") value = "space";
```
ethanaobrien: indentation error

ethanaobrien (changes requested):

data/emulator.js
```diff
@@ -452,18 +452,16 @@ class EmulatorJS {
...
-                    console.log("Translation not found for '"+text+"'. Language set to '"+this.config.language+"'");
-                }
-                return this.config.langJson[text] || text;
+        if (typeof text === "undefined") return;
+        text = text.toString();
+        if (text.includes("EmulatorJS v")) return text;
+        if (this.config.langJson) {
+            if (typeof log === "undefined") log = true;
+            if (!this.config.langJson[text] && log) {
+                if (!EJS_missingLang.includes(text)) EJS_missingLang.push(text);
```
ethanaobrien: Where is EJS_missingLang defined

ethanaobrien (changes requested):

data/emulator.js
allancoding: Where should I define it?

ethanaobrien: On the `this` object within the constructor. Maybe only name it `missingLang`

allancoding: like that?

ethanaobrien: yep, that's good

data/emulator.js
```diff
@@ -2520,9 +2527,9 @@ class EmulatorJS {
         this.elements.parent.appendChild(this.virtualGamepad);
 
         const speedControlButtons = [
-            {"type":"button","text":"Fast","id":"speed-fast","location":"center","left":-35,"top":50,"fontSize":15,"block":true,"input_value":27},
-            {"type":"button","text":"Slow","id":"speed-slow","location":"center","left":95,"top":50,"fontSize":15,"block":true,"input_value":29},
-        ];
+            {"type":"button","text":this.localization("Fast"),"id":"speed-fast","location":"center","left":-35,"top":50,"fontSize":15,"block":true,"input_value":27},
+            {"type":"button","text":this.localization("Slow"),"id":"speed-slow","location":"center","left":95,"top":50,"fontSize":15,"block":true,"input_value":29},
+        ];
```
ethanaobrien: Can you remove these extra spaces, and didn't we move to where the localization is set afterwards?

ethanaobrien (changes requested):

zmarteline:
Ok, we are almost there...
Some issues:
1- buttons and direcionals translations are followed by the word undefined;

2- the  stickers, use the number 1 or -1 from the label,  could that label be ommited so it just appears X and -X? Otherwise we will get as result in the label the sticker information plus the label for regular button 1.

3- keyboard translations are not working.
fails

ethanaobrien:
@allancoding Make sure localization never returns undefined. If it tries to return undefined it should return the original

data/emulator.js
```diff
@@ -2245,18 +2246,24 @@ class EmulatorJS {
...
                     if (this.controls[i][k] && this.controls[i][k].value !== undefined) {
-                        textBox2.value = this.controls[i][k].value;
+                        let value = this.controls[i][k].value.toString();
+                        if (value === " ") value = "space";
+                        textBox2.value = value;
                     }
-                    if (this.controls[i][k] && this.controls[i][k].value2 !== undefined) {
-                        textBox1.value = this.controls[i][k].value2;
+                    if (this.controls[i][k] && this.controls[i][k].value2 !== undefined && this.controls[i][k].value2 !== "") {
+                        let value2 = this.controls[i][k].value2.toString().split(":");
```
ethanaobrien, line 2254: This broke it. It **should not** be assumed that the value has a colon. This is only on axis

data/emulator.js
```diff
@@ -2245,18 +2246,24 @@ class EmulatorJS {
...
                 if (this.controls[i][k] && this.controls[i][k].value) {
-                    textBox2.value = this.controls[i][k].value;
+                    let value = this.controls[i][k].value.toString();
+                    if (value === " ") value = "space";
+                    textBox2.value = value;
                 }
                 if (this.controls[i][k] && this.controls[i][k].value2) {
-                    textBox1.value = "button " + this.controls[i][k].value2;
+                    let value2 = this.controls[i][k].value2.toString().split(":");
+                    textBox1.value = this.localization("button") + " " + this.localization(value2[0]) + ":" + this.localization(value2[1]);
```
ethanaobrien, line 2266: here too

data/emulator.js
allancoding: I will have it check if a : exists

zmarteline:
Let me know when you need me to test again.

allancoding:
@zmarteline I made my changes in #635

zmarteline:
There is no changes regarding labels...

PR #635, opened 2023-08-25
joystick and keyboard localization fix

allancoding:
fixes some things in #630


allancoding:
@zmarteline Can you test this?

data/emulator.js
```diff
@@ -2249,22 +2249,36 @@ class EmulatorJS {
...
                         let value = this.controls[i][k].value.toString();
                         if (value === " ") value = "space";
+                        value = this.localization(value);
                         textBox2.value = value;
                     }
                     if (this.controls[i][k] && this.controls[i][k].value2 !== undefined && this.controls[i][k].value2 !== "") {
-                        let value2 = this.controls[i][k].value2.toString().split(":");
-                        textBox1.value = this.localization(value2[0]) + ":" + this.localization(value2[1]);
+                        let value2 = this.controls[i][k].value2.toString();
+                        if (value2.indexOf(":") !== -1) {
```
ethanaobrien: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

data/emulator.js
```diff
@@ -2249,22 +2249,36 @@ class EmulatorJS {
...
+                            value2 = value2.split(":");
+                            value2 = this.localization(value2[0]) + ":" + this.localization(value2[1])
+                        }else{
```
ethanaobrien: Should be `} else {`

data/emulator.js
```diff
@@ -2249,22 +2249,36 @@ class EmulatorJS {
...
                     let value = this.controls[i][k].value.toString();
                     if (value === " ") value = "space";
+                    value = this.localization(value);
                     textBox2.value = value;
                 }
                 if (this.controls[i][k] && this.controls[i][k].value2) {
-                    let value2 = this.controls[i][k].value2.toString().split(":");
-                    textBox1.value = this.localization("button") + " " + this.localization(value2[0]) + ":" + this.localization(value2[1]);
+                    let value2 = this.controls[i][k].value2.toString();
+                    if (value2.indexOf(":") !== -1) {
```
ethanaobrien: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

data/emulator.js
```diff
@@ -2249,22 +2249,36 @@ class EmulatorJS {
...
+                        value2 = value2.split(":");
+                        value2 = this.localization(value2[0]) + ":" + this.localization(value2[1])
+                    }else{
```
ethanaobrien: Should be `} else {`

ethanaobrien (changes requested):

data/emulator.js
allancoding: Yes lol gotta have those spaces.

data/emulator.js
allancoding: Probably better to use include

zmarteline:
There is no changes regarding labels:
z1z

ethanaobrien:
@zmarteline are you using his branch or still the base emulatorjs repository

allancoding:
My fork is: https://github.com/allancoding/EmulatorJS/tree/main

zmarteline:
Sorry, I was not aware that I should download from a different place. 
Ok,  tested, nicely done! Now the only issue remaining is the conflict caused because the value  "1" from sticker action uses the same label from "button 1
erroz

"

zmarteline:
[pt-BR.json.txt](https://github.com/EmulatorJS/EmulatorJS/files/12442623/pt-BR.json.txt)
Translation file updated with included labels.

allancoding:
@zmarteline Those numbers need to stay a number not text. The file I will use is
[pt-BR.json.txt](https://github.com/EmulatorJS/EmulatorJS/files/12444416/pt-BR.json.txt)
Other parts of the code uses those numbers as numbers and not buttons they are there for the languages that do not use the latin number system.
As in the english translation changing it from 1 to button 1 is not a correct translation.
I did add the other parts tho.
Thanks

allancoding:
The type of controller you are using will designate those values so some controllers will tell you that it is button a or b. Instead of giving a number.

ethanaobrien:
If you translate those numbers in the right place, it should work, shouldn't it. I don't think it uses the innertext value to check anything at all

allancoding:
The controller sends the name as 1 then it should be 1 not button 1 cuz it may not even be a button. And if it translates the 1 to button 1 in the settings if something equals 1 then it will show button 1 and not the value 1.

ethanaobrien:
Can you give examples? All I read was the number 1 like 5 times

zmarteline:
Guys the current translation labels are more than ok. Regarding using the word button was just an attempt to polish it even more, since the buttons return from 3 different gamepads I have are just numbers, as you can see below:
(Buttons  0, 1 and 3, folowed by directionals)

DATAFROG PS4
datafrog

8BITD8 M30
8bitdo

CYBORG X
ctrl 1

Once again, right now is more than ok. Thank a lot for all that effort put into that.

allancoding:
I have an idea that might work I will just have to add something to all the languages.

allancoding:
This is fixed but is continued in #637

PR #651, opened 2023-08-30
neo geo pocket support

n-at:
WASM build of https://github.com/libretro/beetle-ngp-libretro (with the same changes to Makefile as in beetle-pce).

Control scheme and virtual gamepad included.
Fast forward, slow motion and rewind works.


ethanaobrien:
Thank you!

Would you mind also adding this one to the build script, and I created the repo here - https://github.com/EmulatorJS/beetle-ngp-libretro

Thank you!

PR #653, opened 2023-08-31
PC-FX support

n-at:
With WASM build of https://github.com/libretro/beetle-pcfx-libretro, control scheme and virtual gamepad.


ethanaobrien:
Thank you!

PR #654, opened 2023-08-31
update rom file search order

n-at:
Update to ROM file search algorithm.
Prevents cue file creation if single-file image (like iso, cso, ...) is loaded or cue file (or ccd, m3u, ...) already exists.
If none found, creates cue (if core supports it) or loads any file with supported extension.


ethanaobrien:
If I remember correctly, the reason why I made it always generate a cue file is because on the mednafen cores (at least the Playstation one) its case sensitive and path sensitive and as a result is a common user error. Have you checked this?

n-at:
I checked yabause (segaSaturn) - there was issues with generated cue files, this algorithm solved it.

Now checked:
* pcsx_rearmed - works
* mednafen_psx_hw - works
File names inside archive was:
```
Metal Gear Solid (USA) (Disc 1)
├── Metal Gear Solid (USA) (Disc 1).bin
└── Metal Gear Solid (USA) (Disc 1).cue
```
(maybe change file names to something else to test?)
* mednafen_pcfx - works

data/emulator.js
```diff
@@ -890,49 +900,58 @@ class EmulatorJS {
...
-                        if (k === "!!notCompressedData") {
-                            if (needsCue && execFile !== null) {
-                                this.fileName = execFile;
+                    let isoFile = null;
+                    let supportedFile = null;
+                    let cueFile = null;
+                    let selectedCueExt = null;
+                    fileNames.forEach(fileName => {
+                        const ext = fileName.split('.').pop().toLowerCase();
+                        if (supportsExt(ext)) {
```
ethanaobrien: Could we only set this if it's not already defined? (Set it to the first file it finds)

data/emulator.js
```diff
@@ -890,49 +900,58 @@ class EmulatorJS {
...
+                            supportedFile = fileName;
+                        }
+                        if (['iso', 'cso', 'chd', 'elf'].includes(ext)) {
+                            isoFile = fileName;
+                        }
+                        if (['cue', 'ccd', 'toc', 'm3u'].includes(ext)) {
+                            if (this.getCore(true) === 'psx') {
+                                //always prefer m3u files for psx cores
+                                if (selectedCueExt !== 'm3u') {
+                                    if (ext === 'm3u' || cueFile === null) {
```
ethanaobrien: Can we also only set this if it hasn't been set

data/emulator.js
```diff
@@ -890,49 +900,58 @@ class EmulatorJS {
...
+                                        cueFile = fileName;
+                                        selectedCueExt = ext;
+                                    }
+                                }
                             } else {
-                                this.fileName = altName;
+                                //prefer cue or ccd files over toc or m3u
+                                if (!['cue', 'ccd'].includes(selectedCueExt)) {
+                                    if (['cue', 'ccd'].includes(ext) || cueFile === null) {
```
ethanaobrien: And same here

data/emulator.js
```diff
@@ -890,49 +900,58 @@ class EmulatorJS {
...
-                            //always prefer m3u files for psx cores
-                            !(this.getCore(true) === "psx" && ["m3u", "ccd"].includes(this.fileName.split(".").pop())))) {
-                            this.fileName = k;
-                        }
-                        if (needsCue && execFile === null && ["m3u", "cue"].includes(k.split(".").pop().toLowerCase())) {
-                            FS.writeFile("/"+k, resData[k]);
+                    });
+                    if (isoFile !== null && (supportsExt('iso') || supportsExt('cso') || supportsExt('chd') || supportsExt('elf'))) {
+                        this.fileName = isoFile;
+                    } else if (!this.config.disableCue && (supportsExt('cue') || supportsExt('ccd') || supportsExt('toc') || supportsExt('m3u'))) {
```
ethanaobrien: On the pcsx reamed core, I think we should enable `disableCue` by default

ethanaobrien (changes requested):

ethanaobrien:
Hmm, alright. I left a review. Once those things are addressed I'll merge

n-at:
Added changes. 

The only problem with file name case I found, if file structure is like this:
```
METAL Gear Solid (USA) (Disc 1)
├── Metal GEAR Solid (USA) (Disc 1).bin
└── Metal Gear SOLID (USA) (Disc 1).cue
```
and the cue file is:
```
FILE "Metal Gear Solid (USA) (Disc 1).bin" BINARY
  TRACK 01 MODE2/2352
    INDEX 01 00:00:00
```
it will not load. I don't know if it's an issue. This file structure will only work on Windows, Linux/Mac will also not be able to load the file.

data/emulator.js
n-at: It's already doing this:
* if matching file found, takes first
* if m3u file found, takes first and replaces current selected (it it wasn't already m3u)

data/emulator.js
n-at: Also enabled for genesis_plus_gx, picodrive and mednafen_pce. They can load both cue and single-file roms.

ethanaobrien:
>it will not load. I don't know if it's an issue. This file structure will only work on Windows, Linux/Mac will also not be able to load the file.

Not technically. The emscripten fs is case sensitive

PR #657, opened 2023-09-03
ability to hide mute button and volume slider

n-at:
`mute` and `volume` buttons are listed in the documentation https://emulatorjs.org/docs/Options.html#ejs-buttons


ethanaobrien:
Nice catch, thanks

PR #683, opened 2023-09-28
Fix controller issues

### LINKED ISSUE #682: Keyboard key W causing problems with A/B inputs
When holding the W KEY on a keyboard, pressing any other keys mapped to the A/B buttons/input to do C-Pad Down & C-Pad Left.

W + (A) = C-Pad Down
W + (B) = C-Pad Left

allancoding:
Also fixed ad iframe block function.

This fixes #682


data/emulator.js
```diff
@@ -2491,132 +2493,125 @@ class EmulatorJS {
...
+            24: {
+                'value': '1'
+            },
+            25: {
+                'value': '2'
+            },
+            26: {
+                'value': '3'
+            },
+            27: {
```
ethanaobrien: I dont think we should have a mapping for fast foward, slow motion, and rewind. Especially since most keyboards dont really have an "add" or "subtract" button, and space seems like something you would automatically set. I think we should leave this up to the user

data/emulator.js
```diff
@@ -2721,21 +2716,22 @@ class EmulatorJS {
     setupKeys(){
         for (let i=0; i<4; i++) {
             for (let j=0; j<30; j++) {
-                if (this.controls[i][j] && this.controls[i][j].value === this.keyMap[this.keyLookup(this.controls[i][j])]) {
-                    this.controls[i][j].keycode = Number(this.keyLookup(this.controls[i][j]));
+                if (this.controls[i][j] && this.keyMap) {
+                    this.controls[i][j].value = Number(this.keyLookup(this.controls[i][j]));
```
ethanaobrien, line 2714: so... value can be a keycode or a string?

data/emulator.js
```diff
@@ -390,8 +390,11 @@ class EmulatorJS {
         if (del){
             document.querySelector('div[class="ejs_ad_iframe"]').remove();
         }else{
-            document.querySelector('iframe[src="'+this.config.adUrl+'"]').src = url;
+            try{
```
ethanaobrien: Can we fix the spacing here.

- functions should always be `function() {`, with a space between the `()` and `{`.
- if statements should always be `if (condition) {` with a space after the `if`, and between the `)` and the `{`
- else statements should always be `} else {`, with a space both before and after the `else`
- try statements should always be `try {`, with a space before the `{` and after the `try`
- catch statements should always be `} catch(e) {}`, with a space after the first `}` and a space before the `{}`

This is the coding standard for this project, to make the code look clean and not cluttered due to the lack of spaces

ethanaobrien (changes requested):

data/emulator.js
allancoding: no

data/emulator.js
allancoding: Done!

data/emulator.js
allancoding: Done!

ethanaobrien (approved):

PR #716, opened 2023-11-05
screen recording

n-at:
Screen recording from this discussion comment: https://github.com/orgs/EmulatorJS/discussions/658#discussioncomment-6906480 

Will add context menu button to start/stop screen recording.

New config options (with defaults):
```js
window.EJS_screenRecording = {
    width: 800,
    height: 600,
    fps: 30,
    videoBitrate: 2 * 1024 * 1014,
    audioBitrate: 256 * 1024,
};
```

### DIFF
--- a/data/emulator.js
+++ b/data/emulator.js
@@ -1281,6 +1281,27 @@ class EmulatorJS {
             a.click();
             hideMenu();
         });
+
+        let screenMediaRecorder = null;
+        const startScreenRecording = addButton("Start screen recording", false, () => {
+            if (screenMediaRecorder !== null) {
+                screenMediaRecorder.stop();
+            }
+            screenMediaRecorder = this.screenRecord();
+            startScreenRecording.setAttribute("hidden", "hidden");
+            stopScreenRecording.removeAttribute("hidden");
+            hideMenu();
+        });
+        const stopScreenRecording = addButton("Stop screen recording", true, () => {
+            if (screenMediaRecorder !== null) {
+                screenMediaRecorder.stop();
+                screenMediaRecorder = null;
+            }
+            startScreenRecording.removeAttribute("hidden");
+            stopScreenRecording.setAttribute("hidden", "hidden");
+            hideMenu();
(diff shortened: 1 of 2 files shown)


ethanaobrien:
Thank you so much! I've been planning on doing this for a while now. I did have 2 questions. These questions don't really matter, I would merge this pr in its current state.

Do you know if it's possible to detect the emulators' screen resolution instead of guessing/having the site owner set a resolution. I assume this would need to be pulled from the webassembly

Does requestanimationframe know to render the frame at the desired recording fps, or will it render at the fps of the host screen?

Also a feature suggestion I had, that would be nice if you had time. Could you check if the `showSaveFilePicker` api is available, and if so use that so save the screen recording, since it will drastically save memory and will write the data directly to the disk. See the attached file for how to use the api, and again, it's completely fine if you don't have time for this

[message (1).txt](https://github.com/EmulatorJS/EmulatorJS/files/13259919/message.1.txt)

n-at:
1. I could not find a simple way to get real screen resolution of emulator. It seems it's not available in JS for now (need to be exported from RetroArch).
2. `requestAnimationFrame` calls rate should match user's screen refresh rate. Originally I tried `setInterval` with recording FPS intervals but frame pacing in recorded video seemed off. Maybe it's because `requestAnimationFrame` is synchronized with emscripten refresh rate.
3. `showSaveFilePicker` API looks interesting, but it's an experimental feature and currently supported only by chrome-based browsers. Maybe I'll look into this some time later (not in this PR).

ethanaobrien (approved):

ethanaobrien:
Alright, I'll go ahead and merge.

I'd like to eventually record the screen from within RetroArch instead of a canvas, but thats for a later time

With showSaveFilePicker, thats why I'd like to use that if available, and fallback if the api is not supported. This can solve major memory issues with the webpage (we dont want to crash mid-game)

ethanaobrien:
And thank you!

PR #728, opened 2023-12-11
Add button to open context menu

andrigamerita:
In some contexts, it's hard or impossible to open the context menu of the emulator, which is required to access important secondary options such as quick states, screen recording, and software information. Primarily, it's apparently impossible to open the context menu on a mobile web browser, since a long press on the screen is, for good reason, not registered like a right mouse click is on desktop. This commit adds what I think is the ideal solution to the issue: a button on the menu bar/drawer that opens up the context menu, without cluttering it up with multiple options, just a single new one.


allancoding:
I think that the `contextMenu` button should only be visible on mobile devices.

ethanaobrien:
I wouldnt say so, nothing bad with having an extra button, and several desktops might either have a damaged mouse/difficult to use touchpad

allancoding:
Yes that is true. Another button is not too bad.

One more thing, `Context Menu` should be added to the `en.json` file.

ethanaobrien:
>One more thing, `Context Menu` should be added to the `en.json` file.

Yep, should be the only thing this pr needs. I like this pr

ethanaobrien (approved):

ethanaobrien:
@allancoding Please review, and squash and merge if it looks good

allancoding:
@andrigamerita Here are some small tweaks.
Could you center the icon in the middle of the button:
Screenshot 2023-12-11 164231
And maybe center the context menu in the center of the screen:
Screenshot 2023-12-11 164349
Other than those 2 small things it looks good!

allancoding:
Looks Great!

ethanaobrien:
For the future, an approving review should be left before merging

PR #771, opened 2024-02-11
Add hooks for saving and loading save files

gantoine:
This duplicates the behavior for `EJS_onLoadState` and `EJS_onSaveState` but for save files. Would like this for some custom behavior for RomM!


ethanaobrien:
Thanks! Looks good

ethanaobrien (approved):

PR #773, opened 2024-02-11
Use package.json version for ejs version checks

gantoine:
This PR replaces the need to manually bump core version when cores change by relying on the existing version in `package.json`. Comparing versions is done naively against the package.json file hosted in the CDN, which lets us drop `version.json`.

### DIFF
--- a/data/emulator.js
+++ b/data/emulator.js
@@ -1,17 +1,5 @@
 const packageJson = require('../package.json');
 
-// Naive semver comparison that only checks major, minor, and patch
-function compareSemver(currentVersion, otherVersion) {
-    const vcparts = currentVersion.split(".").map(Number);
-    const voparts = otherVersion.split(".").map(Number);
-
-    for (let i = 0; i < 3; i++) {
-        if ((vcparts[i] ?? 0) > (voparts[i] ?? 0)) return -1;
-        if ((vcparts[i] ?? 0) < (voparts[i] ?? 0)) return 1;
-    }
-    return 0;
-}
-
 class EmulatorJS {
     ejs_version = packageJson.version;
     version = packageJson.version; // Keeping this alias for backwards compatibility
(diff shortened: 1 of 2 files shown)


data/emulator.js
```diff
@@ -1,5 +1,21 @@
+const packageJson = require('../package.json');
```
ethanaobrien, line 1: Is the `require` available in the web browser? How new is this api, I'm trying to hold onto chrome 55-70 support for as long as possible which is why there are several apis I could use where I don't

data/emulator.js
```diff
@@ -1,5 +1,21 @@
...
+
+// Naive semver comparison that only checks major, minor, and patch
+function compareSemver(currentVersion, otherVersion) {
```
ethanaobrien: Could we either build this function into the EmulatorJS class or into the check for updates function just to keep organization?

data/version.json
```diff
@@ -1 +0,0 @@
-{ "current_version": 401.1 }
```
ethanaobrien: Might want to keep this and just hold it back, since older clients still need to fetch for updates

ethanaobrien (changes requested):

data/emulator.js
gantoine: i think i meant to use `import` here, which is supported in Chrome 61+
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules
Screenshot 2024-02-12 at 10 15 29 AM

ethanaobrien:
Hmm, @gantoine - After thinking about this for quite a while I've been reminded that the `data` folder is meant to be packaged separately so in some (if not most) cases, this would break version checking + display entirely (+ sometimes cache can be weird and cache one file and not the other), though for the update check that still works perfectly. I do also plan on adding a version check to the cores to make sure they are compatible with emulatorjs (Though this can only be loaded once the core is loaded + ready to play, so its more of a warning than an error). I'd like your thought on this before I make any changes. This is also the last change I have planned for the next release

ethanaobrien:
Going to close and re-work this method

PR #774, opened 2024-02-12
Add support for the SMSPlus-GX core

michael-j-green:
This PR adds the SMSPlus-GX core, which seems to more reliably play Sega Master System and Game Gear ROMs, than the current Genesis core.

This new core will be the default used when the segaMS alias is used.

Also added a query string parameter to enable threads: ``http://url/index.html?threads=1``

Threads will normally be disabled

### DIFF
--- a/index.html
+++ b/index.html
@@ -107,15 +107,25 @@
 
         <script>
             var enableDebug = false;
+            var enableThreads = false;
             const queryString = window.location.search;
             const urlParams = new URLSearchParams(queryString);
             if (urlParams.get('debug') == 1)
                 enableDebug = true;
+
+            if (urlParams.get('threads') == 1)
+                enableThreads = true;
             
             if (enableDebug == true) {
                 console.log("Debug is enabled");
             } else {
                 console.log("Debug is disabled");
+            }
+
+            if (enableThreads == true) {
+                console.log("Threads are enabled");
+            } else {
+                console.log("Threads are disabled");
             }
             
             input.onchange = async () => {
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -107,16 +107,26 @@ <h1>EmulatorJS Demo</h1>
 
         <script>
             var enableDebug = false;
+            var enableThreads = false;
```
ethanaobrien: Could we use `let` here

index.html
```diff
@@ -107,16 +107,26 @@ <h1>EmulatorJS Demo</h1>
...
+            if (urlParams.get('threads') == 1)
+                enableThreads = true;
             
             if (enableDebug == true) {
                 console.log("Debug is enabled");
             } else {
                 console.log("Debug is disabled");
             }
+
+            if (enableThreads == true) {
```
ethanaobrien: No need for the ` == true`

index.html
```diff
@@ -107,16 +107,26 @@ <h1>EmulatorJS Demo</h1>
...
         <script>
             var enableDebug = false;
+            var enableThreads = false;
             const queryString = window.location.search;
             const urlParams = new URLSearchParams(queryString);
             if (urlParams.get('debug') == 1)
                 enableDebug = true;
+
+            if (urlParams.get('threads') == 1)
+                enableThreads = true;
```
ethanaobrien: Can we also assert the existence of `SharedArrayBuffer`

ethanaobrien (changes requested):

index.html
michael-j-green: Both enableDebug and enableThreads have now been changed to let

index.html
michael-j-green: Simplified the code by not doing the same test twice.

ethanaobrien (approved):

PR #776, opened 2024-02-13
Fix build error in npm.yml

allancoding:
Needs a package-lock.json

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -13,6 +13,7 @@ jobs:
         with:
           node-version: '20.x'
           registry-url: 'https://registry.npmjs.org'
+      - run: npm i
       - run: npm ci
       - run: npm publish
         env:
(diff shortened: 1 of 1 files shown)


cohenerickson:
You may also need to change the publish command to be `npm publish --access public`. When publishing packages under an organization name the visibility defaults to private instead of public.

allancoding:
Thanks

ethanaobrien (approved):

cohenerickson:
Is it possible to also get the game cores published to NPM or are there limits on the file sizes?

allancoding:
Right now all the cores are 200+ mb don't want to push this the the main repo. but we could make another npm @emulatorjs/cores

ethanaobrien:
And to add, the 200mb of core files will only get bigger

michael-j-green:
Could the cores be deployed separately? Perhaps on a per platform basis rather than one mega package?

Alternative is a stub package that when executed/installed/etc downloads the files from the CDN based on the package version.

allancoding:
yes that would be nice I will look into adding that to the stable script. I might do it by company so like nintendo cores, then sony cores, etc

PR #787, opened 2024-03-01
[Request] Display the file name of the ROM/disk when using M3U lists

### LINKED ISSUE #714: [Request] Display the file name of the ROM/disk when using M3U lists
At present the disk switcher only says "Disk 1", "Disk 2", etc. Thinking this could be confusing as double sided disks are labeled the same way with Disk 1 Side B labeled "Disk 2".

Listing the file name would make it easier to identify which disk the user is changing to.

The M3U spec in the libretro documentation (see here for Amiga, but it's the same for other cores: https://docs.libretro.com/library/puae/#m3u-and-disk-control) also includes support for arbitrarily naming disks - for example "Program" disk, or "Data" disk, by putting the name after the file name separated by a |

From the RetroArch docs:
By default, RetroArch will display the filename (without extension) of each M3U entry when selecting a disk via the Current Disc Index drop-down menu. Custom display labels may be set for each disk using the syntax: ```DISK_FILE|DISK_LABEL```. For example, the following M3U file:

```Valhalla & the Fortress of Eve.m3u```

```
Valhalla & the Fortress of Eve_Disk1.adf|Game Disk
Valhalla & the Fortress of Eve_Disk2.adf|Data Disk
Valhalla & the Fortress of Eve_Disk3.adf|Level 1 Disk
Valhalla & the Fortress of Eve_Disk4.adf|Level 2 Disk
Valhalla & the Fortress of Eve_Disk5.adf|Level 3 Disk
Valhalla & the Fortress of Eve_Disk6.adf|Level 4 Disk
```

...will be shown in RetroArch's disk selection menu as:

```
1: Game Disk
2: Data Disk
3: Level 1 Disk
4: Level 2 Disk
5: Level 3 Disk
6: Level 4 Disk
```

If ```DISK_LABEL``` is intentionally left blank (i.e. ```DISK_FILE|```) then only the disk index will be displayed.

### PR DESCRIPTION
**Pull request by michael-j-green**
When using M3U files, RetroArch supports displaying a string from the M3U file as the disk name.

For example:
If an M3U file like below is provided
```
Summer Games 2 (1985)(Epyx)(Side A)[cr CRM].d64
Summer Games 2 (1985)(Epyx)(Side B)[cr CRM].d64
```

Then the following will be displayed:
Screenshot 2024-03-01 at 9 48 42 pm

Adding a pipe character after each line can be used to set a custom name for the disk:
```
Summer Games 2 (1985)(Epyx)(Side A)[cr CRM].d64|Side A
Summer Games 2 (1985)(Epyx)(Side B)[cr CRM].d64|Side B
```

The following will be displayed:
Screenshot 2024-03-01 at 9 50 28 pm

Closes #714

### DIFF
--- a/data/emulator.js
+++ b/data/emulator.js
@@ -1124,6 +1124,11 @@ class EmulatorJS {
             }
             this.Module.resumeMainLoop();
             this.checkSupportedOpts();
+            this.setupDisksMenu();
+            // hide the disks menu if the disk count is not greater than 1
+            if (!(this.gameManager.getDiskCount() > 1)) {
+                this.diskParent.style.display = 'none';
+            }
             this.setupSettingsMenu();
             this.loadSettings();
             this.updateCheatUI();
@@ -1575,7 +1580,7 @@ class EmulatorJS {
         let timeout = null;
         let ignoreEvents = false;
         const hide = () => {
-            if (this.paused || this.settingsMenuOpen) return;
+            if (this.paused || this.settingsMenuOpen || this.disksMenuOpen) return;
             this.elements.menu.classList.add("ejs_menu_bar_hidden");
         }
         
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by michael-j-green** [comment]
Feel free to mock my JS 🤣

**Comment by michael-j-green** [comment]
I would really love to be able to spin the disk submenu out to it's own toolbar button that's only visible when multiple disks are available. This would make swapping disks much easier with fewer clicks, but that looks like it's going to be a pain to implement.

File: data/emulator.js [hunk 1]
```diff
@@ -4060,8 +4060,29 @@ class EmulatorJS {
         
         if (this.gameManager.getDiskCount() > 1) {
             const diskLabels = {};
+            let isM3U = false;
+            let disks = {};
+            if (this.fileName.split(".").pop() == "m3u") {
```
ethanaobrien: In JavaScript, we should use 3 equal signs. Why JavaScript does this? Who knows

data/emulator.js
```diff
@@ -4060,8 +4060,29 @@ class EmulatorJS {
...
+                disks = FS.readFile(this.fileName, { encoding: 'utf8' }).split("\n");
+                isM3U = true;
+            }
             for (let i=0; i<this.gameManager.getDiskCount(); i++) {
-                diskLabels[i.toString()] = "Disk "+(i+1);
+                // default if not an m3u loaded rom is "Disk x"
+                // if m3u, then use the file name without the extension
+                // if m3u, and contains a |, then use the string after the | as the disk label
+                if (isM3U == false) {
```
ethanaobrien: Same here, though we could just say if !m3u

data/emulator.js
```diff
@@ -4060,8 +4060,29 @@ class EmulatorJS {
         
         if (this.gameManager.getDiskCount() > 1) {
             const diskLabels = {};
+            let isM3U = false;
+            let disks = {};
+            if (this.fileName.split(".").pop() == "m3u") {
+                disks = FS.readFile(this.fileName, { encoding: 'utf8' }).split("\n");
```
ethanaobrien: Can we access FS though this.Module.FS

ethanaobrien (dismissed):

data/emulator.js
michael-j-green: I looked it up ```===``` is a strict equality. You need to do this in JS because JS isn't strongly typed like c# or other languages are.

ethanaobrien: That's it, I can never actually remember what it means so I just never use == in JavaScript

data/emulator.js
ethanaobrien: this.gameManager.Module.FS*

michael-j-green:
@ethanaobrien; If you're good to go on this PR, then I'm happy too. In all my testing the disk naming adhere's to the standard from the M3U sections in the libretro documentation, with appropriate fall back to the disk file name (sans file extension).

The menu implemented in PR #790 also works correctly.

Edit: the code also only kicks in if it was an M3U file that was provided. Other methods of populating this menu fall back to the original method using the "Disk X" string - though I can't imagine what other methods there are...

@allancoding; would you be able to put a note in here if you're happy with the styling of #790 (this has been merged into this PR)?

allancoding:
Yeah im very happy with it.

allancoding (approved):
Looks great!

data/emulator.js
```diff
@@ -1880,6 +1885,41 @@ class EmulatorJS {
...
+            diskButton[2].classList.toggle("ejs_disks_text", this.disksMenuOpen);
+            this.disksMenu.style.display = "none";
+        }
+        this.addEventListener(this.elements.parent, "mousedown touchstart", (e) => {
+            if (this.isChild(this.disksMenu, e.target)) return;
+            if (e.pointerType === "touch") return;
+            if (e.target === diskButton[0] || e.target === diskButton[2]) return;
+            this.closeDisksMenu();
+        })
+        this.addEventListener(this.canvas, "click", (e) => {
```
ethanaobrien: Is this new?

michael-j-green: Yes... it's there to close the disks menu.

michael-j-green: The disks menu is a copy/paste of the settings menu - just with the word "Settings" replaced with "Disks" :D

ethanaobrien: It appears to also attempt to lock the pointer. Did you remove the trigger from the other place?

ethanaobrien: > The disks menu is a copy/paste of the settings menu - just with the word "Settings" replaced with "Disks"

Oh lol... didn't think I made it that modular lol nice

michael-j-green: > It appears to also attempt to lock the pointer. Did you remove the trigger from the other place?

It shouldn't... I reverted the change in this PR back to the original behaviour.
Screenshot 2024-03-04 at 12 10 03 pm
I'll double check what's been submitted to make sure though.

michael-j-green: > > The disks menu is a copy/paste of the settings menu - just with the word "Settings" replaced with "Disks"
> 
> Oh lol... didn't think I made it that modular lol nice

haha... there is probably an opportunity for refactoring down the track to make menu's from the toolbar button upwards, but for now, this'll do.

michael-j-green: Yup... nds is back as a condition for pointer capture.
Screenshot 2024-03-04 at 12 12 15 pm

ethanaobrien: I'm just confused why all the lines show as added, did your change get pushed to upstream? I never saw any pointer lock function calls removed

michael-j-green: Ah... that was my bad... I accidentally committed the pointer lock changes to wrong branch. Didn't notice until I was already deep in it, so just reverted it manually - that's why it shows as added.

ethanaobrien: Ah ok, so that needs to be changed or is this ready to merge?

michael-j-green: Ready to merge. When I run this code locally, it doesn't capture my pointer, and behaves as the current production version does.

ethanaobrien: It appears to be duplicated?

https://github.com/EmulatorJS/EmulatorJS/blob/e11b5fc7748463d1c71e0e1ee57fa104a41253de/data/emulator.js#L1911

https://github.com/EmulatorJS/EmulatorJS/blob/e11b5fc7748463d1c71e0e1ee57fa104a41253de/data/emulator.js#L1945

michael-j-green: You're right... I think I included it there because it needed to be there for each menu. I'll strip it out and test.

michael-j-green: Yup... works fine with just the one.

This is why you don't code at 2:30am folks :D

michael-j-green: That looks better...
Screenshot 2024-03-04 at 12 30 22 pm

michael-j-green: I put a new line in between to make it clear it's got nothing to do with the settings menu.

ethanaobrien (approved):

PR #789, opened 2024-03-01
Disables the right click menu from displaying

michael-j-green:
The right click context menu that appears obscures the game canvas when playing games that use a mouse and right clicks.

The menu is still displayed when clicking the burger icon in the toolbar, so functionality is not lost - just moved.


allancoding:
I think that the right click menu need to stay on as default and just make an option to disable it/maybe join it with the show menu button or something.

ethanaobrien:
I don't really think we should disable the right click menu. There's an option to disable it, and if the cursor is locked to the screen it shouldn't trigger

michael-j-green:
Fair :)

I'll close it off.

PR #802, opened 2024-03-16
Extended shader configuration format and custom shaders

n-at:
Added extended shader configuration format. It allows to define shaders consisting of multiple files. Previous format also works.

Added `EJS_shader` configuration for custom shaders:

```js
window.EJS_shaders = {
    //https://github.com/libretro/glsl-shaders/blob/master/anti-aliasing/fxaa.glslp
    "fxaa": {
        "shader": {
            "type": "text",
            "value": "shaders = 1\n\nshader0 = fxaa.glsl\nfilter_linear0 = true\n",
        },
        "resources": [
            {
                "name": "fxaa.glsl",
                "type": "base64",
                "value": "...BASE64 string...",
            },
        ],
    },
};
```

Added SABR and CRT beam shaders.


ethanaobrien:
Just a note: I won’t be able to review for about a week

ethanaobrien:
Finally got around to reviewing - thank you!

ethanaobrien (approved):

PR #855, opened 2024-07-14
Fixed the error of the "foreach" function in the browser.

ZhaoTonggang:
The objects traversed by the “foreach” method should be arrays, and the “Array.from()” method should be used to convert the traversal objects to arrays, ensuring that they run in some more restrictive browsers.

### DIFF
--- a/data/gamepad.js
+++ b/data/gamepad.js
@@ -37,7 +37,7 @@ class GamepadHandler {
         this.timeout = setTimeout(this.loop.bind(this), 10);
     }
     updateGamepadState() {
-        const gamepads = this.getGamepads();
+        const gamepads = Array.from(this.getGamepads());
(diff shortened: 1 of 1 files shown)


ethanaobrien:
Interesting issue, thanks for figuring that out

Could you open a pr in this repo? https://github.com/ethanaobrien/Gamepad

if not, that’s all good. I’d just prefer you do it to keep commit authors even across repos

ethanaobrien (approved):

ZhaoTonggang:
> Interesting issue, thanks for figuring that out
> 
> Could you open a pr in this repo? https://github.com/ethanaobrien/Gamepad
> 
> if not, that’s all good. I’d just prefer you do it to keep commit authors even across repos

Yes, I would be more than happy to do so.

Issue #878, opened 2024-08-18
Add same_cdi core


pastisme:
Hello,

I am trying to integrate [SAME_CDI](https://github.com/libretro/same_cdi) into emulatorJS.
I managed to get it running.

![image](https://github.com/user-attachments/assets/ec1bc453-78b7-460c-a91f-23653ed950c2)

I encountered two issues and the way I worked around them are not very future proof, I was wondering if you have been there before and have any insights or better practice recommendations.

### liblzma conflict
**Problem**: both [same_cdi](https://github.com/libretro/same_cdi/tree/master/3rdparty/lzma/C) and [Retroarch](https://github.com/EmulatorJS/RetroArch/tree/master/deps/7zip) provide their own different version of liblzma.
**Workaround**: I moved over Retroarch files for that lib to same_cdi and adjusted build files.
Any better way to do it ? Is there a Makefile flag to prevent lzma linking when linking the core .bc file with Retroarch ?

### emscripten and exception handling
**Problem**: the code that emulates the m68k cpu relies on C++ exceptions in its logic. [emscripten](https://emscripten.org/docs/porting/exceptions.html) have them unsupported by default for performance purposes.
**Workaround**: I use EXCEPTION_CATCHING_ALLOWED to allow that one [method](https://github.com/libretro/same_cdi/blob/master/src/devices/cpu/m68000/m68kcpu.cpp#L822) to work with exception. It is a class method and it seems that EXCEPTION_CATCHING_ALLOWED does not accept unmangled C++ method. This will break if the compiler decides to change its mangling format.
Do you know a better way ?

Thanks

ethanaobrien:
>liblzma conflict

No better way to do that afaik. I've also had to do some workarounds with various cores similar to what you did, though never had to replace the entire library

>emscripten and exception handling

I know of no alternative way to enable exceptions, especially since I havent played with C++ with emscripten at all. `EXCEPTION_CATCHING_ALLOWED` is probably the "safest" way to ensure it works

Please send a PR to https://github.com/EmulatorJS/same_cdi and https://github.com/EmulatorJS/build so the commit history shows you as a contributor

pastisme:
Can do.

One other issue, this time on the JS side, [same_cdi](https://docs.libretro.com/library/same_cdi/#bios) requires the zip file of the bios. But EmulatorJS seems to always unzip any attached zip file.

Is changing the logic in the core the only way ?

ethanaobrien:
[...]

PR #881, opened 2024-08-21
Let same_cdi core handle zipped bios file directly

pastisme:
#878 
[same_cdi requires a zip file for the bios](https://docs.libretro.com/library/same_cdi/#bios), the core will extract the content into a specific directory.
The resulting file system structure is needed for the core to function properly.

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -782,6 +782,11 @@ class EmulatorJS {
             }
             this.textElem.innerText = this.localization("Download Game BIOS");
             const gotBios = (data) => {
+                if (this.getCore() === "same_cdi") {
+                    this.gameManager.FS.writeFile(this.config.biosUrl.split('/').pop().split("#")[0].split("?")[0], new Uint8Array(data));
+                    resolve();
+                    return;
+                }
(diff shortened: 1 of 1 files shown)


data/src/emulator.js
```diff
@@ -782,17 +782,22 @@ class EmulatorJS {
...
-                            break;
-                        }
-                        if (k.endsWith('/')) continue;
-                        this.gameManager.FS.writeFile("/" + k.split('/').pop(), data[k]);
-                    }
+                if (this.getCore() === "same_cdi") {
+                    this.gameManager.FS.writeFile(this.config.biosUrl.split('/').pop().split("#")[0].split("?")[0], new Uint8Array(data));
                     resolve();
-                })
+                } else {
```
ethanaobrien: No need to have an else statement here, we can return from inside the if statement since it handles the rest of the function

similar to how downloading the rom works: https://github.com/EmulatorJS/EmulatorJS/blob/9598c58cf8089833b3dfb0bd193a58ba4b77b87f/data/src/emulator.js#L848

ethanaobrien (changes requested):

ethanaobrien:
Awesome! Thanks for the pr!

ethanaobrien (approved):

PR #903, opened 2024-09-20
use Module.AL instead of window.AL

### LINKED ISSUE #879: [Bug] Volume Slider Doesnt Change Volume
In 4.1.1, the volume slider doesnt actually change the volume level. There is nothing output in the console logs. I have tested on Firefox and Chrome from Windows and Safari in iOS.

n-at:
Will resolve #879 after cores will be built with this update: https://github.com/EmulatorJS/RetroArch/pull/23

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -1083,8 +1083,8 @@ class EmulatorJS {
             let state = "suspended";
             let popup;
             while (state === "suspended") {
-                if (!window.AL) return;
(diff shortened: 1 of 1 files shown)


ethanaobrien:
Thank you for the fix!

ethanaobrien (approved):

pokemonsilvergoldred:
yes! =D

PR #923, opened 2024-12-21
Fix SharedArrayBuffer check & video rotation without reload 

allancoding:
- video rotation without reload 
- Fix SharedArrayBuffer

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -3986,6 +3986,8 @@ class EmulatorJS {
             }
         } else if (option === "vsync") {
             this.gameManager.setVSync(value === "enabled");
+        } else if (option === "videoRotation") {
+            this.gameManager.setVideoRotation(value);
         }
     }
     menuOptionChanged(option, value) {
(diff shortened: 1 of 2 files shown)


data/src/emulator.js
```diff
@@ -3986,14 +3986,19 @@ class EmulatorJS {
...
+            this.gameManager.setVideoRotation(value);
         }
     }
     menuOptionChanged(option, value) {
         this.saveSettings();
         if (this.debug) console.log(option, value);
         if (!this.gameManager) return;
         this.handleSpecialOptions(option, value);
-        this.gameManager.setVariable(option, value);
+        const notCoreOption = [ "webgl2Enabled", "videoRotation"];
```
ethanaobrien: What is this for

ethanaobrien (changes requested):

data/src/emulator.js
allancoding: There is no need to send invalid core options to retroarch

ethanaobrien: I don't think we should check for this, There are many more "invalid" core options than just these 2, and we're trying to move away from magic variables. The setVariable function also properly handles unknown core options

ethanaobrien (approved):

PR #924, opened 2024-12-21
Use pinned ejs_version when fetching cores/version

gantoine:
Instead of fetching the stable release, which could get out of sync with new release, fetch the release defined in the script version as `this.ejs_version`.


ethanaobrien:
Will need to wait until after @allancoding adds the ability to the cdn to symlink https://cdn.emulatorjs.org/4.2.0/ to https://cdn.emulatorjs.org/latest/

ethanaobrien (approved):

ethanaobrien:
I just caught this and will fix, but the update check should always pull from stable...
{9C42E082-9102-4A10-8132-6A6C62813023}

Issue #249, opened 2022-05-20
Web demo bios upload


ghost:
Some emulators need bios to run correctly, but the default home page only has an option to load the game rom, I hope you can fix that.

ethanaobrien:
What exactly do you mean? For example, [the nes readme](https://github.com/ethanaobrien/emulatorjs/blob/main/docs/Systems/NES-Famicom.md) has the option `EJS_biosUrl = '';`. You set this to the location of the bios file and can be set on any system (doesnt mean it would do something for them all), this is how you reference the bios. Is this what youre asking?

ghost:
> What exactly do you mean? For example, [the nes readme](https://github.com/ethanaobrien/emulatorjs/blob/main/docs/Systems/NES-Famicom.md) has the option `EJS_biosUrl = '';`. You set this to the location of the bios file and can be set on any system (doesnt mean it would do something for them all), this is how you reference the bios. Is this what youre asking?

As you know, playing games on the Turbografx-16 CD system requires a bios in order to load the CD games correctly, but index.html does not contain this, only a drag and drop game to load option.
![Screenshot_2022-05-22-10-41-41-906_org mozilla fenix](https://user-images.githubusercontent.com/32984303/169675911-7ce28798-76fc-4945-b465-fa3aaf922885.png)

ethanaobrien:
Ah ok got it, thank you

ethanaobrien:
I've decided to not do this, the demo is supposed to be for people to try it out, not to actively play games on it.

PR #954, opened 2025-02-15
Add support for Arcade BIOS files

pjft:
The default logic decompresses the BIOS file if it's a zip. 

On arcade systems, it is quite normal to have BIOS files _being_ an actual zip file. Decompressing it makes the game loading fail, as it expects to find a file named like the one we uploaded. For instance, neogeo.zip. 

This approach stores the zip file before decompressing, if it's an arcade or mame system, and then proceeds to decompressing - in case it's a zip with multiple zip files for some more complex systems with multiple drivers. 

The proper solution would be for us to allow for multiple BIOS attachments, but this compromise would work.

Please check if it doesn't break anything, but this fixes fbneo and mame2003 BIOS uploads. Tested with NeoGeo games and PGM games. If you right now try to load NeoGeo games or PGM games, for instance, it'll fail to do so - unless you add the actual BIOS files inside the rom zip file, but it doesn't do anything with the BIOS file in arcade systems.

Thanks and keep up the great work!

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -680,7 +680,7 @@ class EmulatorJS {
                 return resolve(assetUrl);
             }
             const gotData = async (input) => {
-                if (['arcade', 'mame'].includes(this.getCore(true))) {
+                if (window.EJS_dontExtractBIOS === true) {
                     this.gameManager.FS.writeFile(assetUrl, new Uint8Array(input));
                     return resolve(assetUrl);
                 }
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -680,7 +680,7 @@ class EmulatorJS {
                 return resolve(assetUrl);
             }
             const gotData = async (input) => {
-                if (window.EJS_dontExtractBIOS === true) {
+                if (this.config.dontExtractBIOS === true) {
                     this.gameManager.FS.writeFile(assetUrl, new Uint8Array(input));
                     return resolve(assetUrl);
                 }
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -682,6 +682,7 @@ class EmulatorJS {
             const gotData = async (input) => {
                 if (['arcade', 'mame'].includes(this.getCore(true))) {
                     this.gameManager.FS.writeFile(assetUrl, new Uint8Array(input));
+                    return resolve(assetUrl);
                 }
                 const data = await this.checkCompression(new Uint8Array(input), decompressProgressMessage);
                 for (const k in data) {
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -680,6 +680,9 @@ class EmulatorJS {
                 return resolve(assetUrl);
             }
             const gotData = async (input) => {
+                if (['arcade', 'mame'].includes(this.getCore(true))) {
+                    this.gameManager.FS.writeFile(assetUrl, new Uint8Array(input));
+                }
                 const data = await this.checkCompression(new Uint8Array(input), decompressProgressMessage);
                 for (const k in data) {
                     const coreFilename = "/"+this.fileName;
--- a/data/loader.js
+++ b/data/loader.js
@@ -115,6 +115,7 @@
     config.softLoad = window.EJS_softLoad;
     config.screenRecording = window.EJS_screenRecording;
     config.externalFiles = window.EJS_externalFiles;
+    config.dontExtractBIOS = window.EJS_dontExtractBIOS;
     config.disableDatabases = window.EJS_disableDatabases;
     config.disableLocalStorage = window.EJS_disableLocalStorage;
     config.forceLegacyCores = window.EJS_forceLegacyCores;


data/src/emulator.js
```diff
@@ -680,6 +680,9 @@ class EmulatorJS {
                 return resolve(assetUrl);
             }
             const gotData = async (input) => {
+                if (['arcade', 'mame'].includes(this.getCore(true))) {
+                    this.gameManager.FS.writeFile(assetUrl, new Uint8Array(input));
```
ethanaobrien, line 684: We should return here - no need to decompress the file

ethanaobrien (changes requested):

data/src/emulator.js
pjft: Well, it's certainly the technically correct option. 
But that's what I was saying: there might be scenarios where you will need multiple BIOS files, and you currently cannot attach them. Furthermore, this change may break the behavior that some users might already be expecting, so I chose to be conservative in my proposal.
But I agree - in the "expected" scenario, you won't need this (especially if there's an option to add more than one file, which currently doesn't exist). But if a game needs multiple BIOS, you won't have other ways to add them to the emulator's data folder.
Nonetheless, I don't know of any such example on the top of my head, so if you approve of us returning here, I'll make the change. Let me know afterwards if I shall reverse it.
Thanks for the prompt review.

michael-j-green:
I don't agree that this is the correct method. I've had instances where I've attached arcade BIOS files in a zip to EJS when using MAME or FBNEO and it's worked correctly.

There's also the impact of changing behaviour that users expect to consider. This is a user un-friendly solution to the problem.

A better solution might be to create an extra option called "EJS_DontExtractBIOS" to optionally prevent unzipping.

pjft:
> There's also the impact of changing behaviour that users expect to consider. This is a user un-friendly solution to the problem.

I agree with this angle, hence my first proposal. 

> I've had instances where I've attached arcade BIOS files in a zip to EJS when using MAME or FBNEO and it's worked correctly.

I would expect that to have been the case in the past, but it isn't what's happening right now and I was unable to replicate any such working instance.

For what it's worth, the current logic was entirely rewritten two months ago, so that might have had an impact here.

https://github.com/EmulatorJS/EmulatorJS/commit/3acb06a0f0aebbe282f2311287b9e1c2c5671717

Nonetheless, I appreciate you all looking into this. The main object of the issue is that right now it doesn't work because we are decompressing the BIOS zip files on arcade emulators, which goes against what the emulators expect.

I have no issues with this approach being scrapped and something else being put together to fix this. After all, you all know the code base and architecture much better than I do - I just looked at the symptom and addressed the cause where I was able to adjust it.

Happy to adjust, or retract this proposal knowing something else will be put in place.

Have a great weekend.

michael-j-green:
All good! Apologies if I came off as dismissive, I was just trying to put in my 2c. All PR's are welcome :)

I'll leave it to @ethanaobrien to make the call though.

pjft:
> All good! Apologies if I came off as dismissive, I was just trying to put in my 2c. All PR's are welcome :)
> 
> I'll leave it to @ethanaobrien to make the call though.

Not at all, but I appreciate the response! I was just keen on solving this for everyone :)

Cheers, and we'll wait for Ethan then.

data/src/emulator.js
ethanaobrien: This is a good point, and makes me reconsider what approach we should take. @michael-j-green proposed  `EJS_dontExtractBIOS` option which may work the best to prevent another breakage. I don't entirely *like* adding more options because it feels like it bloats the codebase a bit each time, though this method may work the best. If it was only for bios files, though, we should also make sure `type === "bios"`

pjft: I can see that being an option, no pun intended. I do need to ask, though: in general, I imagine a user would only not want to extract bios on arcade systems, just because their emulators take zip files. On every other system you would want to unzip zip files, I believe?

If that is the case, then we'd also want to keep the arcade check, to prevent folks from causing issues in other systems, so that option would only work for arcade bios. And be "on" by default. We're only leaving it as an option to accommodate either folks who have already worked around it or expect it to work as such. 

Happy either way, let me know what you all think. 

@michael-j-green tagging you here in case you don't get thread responses, hope that's alright. 

It's late where I'm at, I'll check in tomorrow. 

Thanks all.

michael-j-green: Going to let @ethanaobrien decide this one :)

I tend to err on the side of not changing current behaviour unless absolutely necessary, so I feel like only not unzipping when requested.

If I remember correctly, RetroArch just needs the files in a specific directory that it'll look in for the BIOS files. You could use the EJS_externalFiles option (https://emulatorjs.org/docs/options#ejs_externalfiles) which is an array to upload files to the directory of your choice - in this case, the BIOS directory where RetroArch will look for them when started.

pjft: Got it. Yes, I believe that's correct - though these particular emulators will look for the zipped files in that directory. 

With this in mind, I side with you in that adding an option as the least risky approach. I share your concern with breaking changes:)

I'll wait for Ethan as well.

pjft: ... Just to be clear, though, since we're also hesitating about adding the option: the reason we're considering adding the option is because we feel that both writing the zip _and_ decompressing as well isn't a good compromise to both fix it and keep the current behavior compatibility? The main trade-off is that it uses up slightly more browser storage, but it ends up being transparent for the user.

michael-j-green: This is why I suggested the EJS_externalFiles option. It's purpose is to allow you upload files to anywhere in the EJS file system.

I think that this one should be checked first to make sure it doesn't satisfy the requirement first before we add extra options or change code.

ethanaobrien: > ... Just to be clear, though, since we're also hesitating about adding the option: the reason we're considering adding the option is because we feel that both writing the zip _and_ decompressing as well isn't a good compromise to both fix it and keep the current behavior compatibility? The main trade-off is that it uses up slightly more browser storage, but it ends up being transparent for the user.

Yes, its not about browser storage but memory usage. The emscripten tempfs is entirely in memory, so more files leads to more memory used

ethanaobrien: @allancoding

pjft:
Hi team.

I've been out for work reasons, and honestly, I have mixed feelings about progressing with this PR at this stage :) 

First: I tend to support with the `EJS_externalFiles` suggestion, though it's not clear to me the extent of the required changes for that to happen. Would we include a separate selector for those (multiple) files then in the current index.html file or something?

Second: the current setup I have now, to work around the current limitation, makes it simpler to navigate the multiple BIOS requirements for arcade games. I just have a single MAME2003-FBNeo zip file with all the BIOS they support (it's less than 10MB), and just add it blindly to any arcade game I run. This simplifies the process significantly, as I do not need to recall what BIOS is required for which rom I want to play. In my regular RetroPie setup I just have them all in the folder and RetroArch just finds the files in the folder, but in this case since it'll read the files from tmpfs, it's just easier to always send them all there.

As such, I'm sharing this here to seek some perspective or more detailed guidance. 

The truth is that for my particular use case, I have a workaround.
Second, if this hasn't come up yet from other users, either they have their own ways to make this work, or they just don't use the arcade emulators for ROMs that require BIOS files.
Last, even with the proposed solutions, the BIOS management for arcade games will still be cumbersome if we have to select individual BIOS files for each game - my particular workaround actually makes that a non-issue, and likely the Rom sets people are using might already have the BIOS inside the games' zip files if they're not running into this.

So I am left wondering whether this is something that I should just retract and let it be, as I'm not thrilled with adding something that will just make things more complex with little benefit for the user base, and that could potentially be solved with a documentation note saying that for arcade games, just zip the BIOS file.

Happy to hear your thoughts, and I appreciate the time you all took.

ethanaobrien:
>I tend to support with the EJS_externalFiles suggestion, though it's not clear to me the extent of the required changes for that to happen

See https://emulatorjs.org/docs/options#ejs_externalfiles

> Would we include a separate selector for those (multiple) files then in the current index.html file or something?

I don't plan to add a seperate selection for BIOS to the demo page, see https://github.com/EmulatorJS/EmulatorJS/issues/249#issuecomment-1214262501. On the demo page, users can compress their ROM and their BIOS together if needed.


I think the simplest solution is the `EJS_dontExtractBIOS` variable. You'll just set it if you dont want the BIOS to extract itself. It's the exact same thing you have now but instead of using hardcoded "if this system, or this system" it would be "if the admin wants it"

pjft:
Got it - that makes it clearer. 

I think I made the change you asked for - I can't see it reflected in the conversation thread, though, but let me know if this is what you had in mind.

Thanks for the patience.

ethanaobrien:
This looks good! Thank you! Last comment I have is we should pull config from `this.config.dontExtractBIOS` like everything else: https://github.com/pjft/EmulatorJS/blob/cc34eb7b391bc5a683cd0c32cb2ac120de7186e7/data/loader.js#L81

pjft:
Thanks. Apologies, I missed that - I saw a few references to window.EJS_xxx and thought that was the proper approach.

I hope this is appropriate now, but if not let me know.

Thanks and have a great weekend!

ethanaobrien (approved):

ethanaobrien:
Yeah.. there are a few usages that have slipped through reviews. I'll go back and change them (eventually). Thanks for the contribution!

PR #982, opened 2025-04-06
Update ja-JA.json

noel-forester:
There were some parts of the Japanese that seemed strange, so they were corrected.


ethanaobrien (approved):
Thank you

PR #992, opened 2025-04-19
Enhance toolbar customisation options

### LINKED ISSUE #981: Extend EJS_Buttons
Suggest a new structure similar to:

```
EJS_Buttons = {
   <buttonName1>: false,
   <buttonName2>: {
      icon: <url to new icon>,
      callback: <javascript function>,
      displayName: <button display name on hover>
   }
}
```

Existing names would be reserved names:
* playPause
* restart
* mute
* settings
* fullscreen
* saveState
* loadState
* screenRecord
* gamepad
* cheat
* volume
* saveSavFiles
* loadSavFiles
* quickSave
* quickLoad
* screenshot
* cacheManager
* exitEmulation

The toolbar would observe the following rules:
* If the entry is omitted from the list altogether, then no change
* If the entry has the value false, then the button is hidden (as per current behaviour)
* If the entry has an array for the value, then the icon and callback attributes become available. Omission of any attribute for reserved name buttons passes through the existing value (for example: if there no `icon` attribute for the `playPause` button means the button will use the existing icon), however custom buttons _must_ use at least the `icon` and `callback` attributes in order for the button to be displayed in the toolbar. The `callback` attribute _does_ note replace the existing behaviour for a reserved button, but can be used as an "event" for the page to detect when a button is pressed.

The above provides for the dev to create custom buttons with any name; however to prevent clashes in the future a potential option is to create a list of say 5 custom button names like `customBtn1`, `customBtn2`, etc.

michael-j-green:
See issue: https://github.com/EmulatorJS/EmulatorJS/issues/981

This change enhances the `EJS_Buttons` option to allow more configuration options than just hiding or showing the button.

Defaults are set for each button in `loader.js`, which are then overriden when the `EJS_Buttons` variable is read.

The logic is as follows:
- if the provided value for the button is a boolean, then the button will simply be hidden or shown as per existing behaviour
- if an array is provided, the button will be configured as per the array. Missing options from the array are simply ignored in favour of the defaults.

An example array structure is as follows:
```javascript
{
    buttonName: {
        visible: true,
        icon: '<svg xml />',
        displayName: 'My Button',
        callback: () => {
            alert("A button was pressed");
        }
    }
}
```

**Note**: This change modifies some internal behaviour. If you're leveraging internal API's regarding button management, please ensure you review your code.

Website documentation updated in Pull Request https://github.com/EmulatorJS/emulatorjs.org/pull/55


michael-j-green:
PR is now functionally complete.

Awaiting guidance from @ethanaobrien as to the suitability of the implementation, if things need to be moved, etc before undoing the changes made by the VSCode formatter (this will need to be done before the PR can be merged!).

data/src/emulator.js
```diff
@@ -1697,13 +1692,18 @@ class EmulatorJS {
...
-        }, async () => {
+        const netplay = addButton(this.config.buttonOpts.netplay, async () => {
             this.openNetplayMenu();
         });
+        if (this.config.buttonOpts.customBtn1 && this.config.buttonOpts.customBtn1.visible === true && this.config.buttonOpts.customBtn1.icon !== undefined && this.config.buttonOpts.customBtn1.callback !== undefined) {
```
allancoding: Could we make it an array so that you are not limited to 3 buttons?

michael-j-green: We can... this is just a proof of concept.

I didn't want to get too in the weeds until @ethanaobrien had a chance to review.

data/loader.js
```diff
@@ -1,4 +1,4 @@
-(async function() {
+(async function () {
```
ethanaobrien: Please tell your IDE to not format changes for this PR; It makes it very difficult to review

data/loader.js
```diff
@@ -91,7 +91,214 @@
     config.adSize = window.EJS_AdSize;
     config.alignStartButton = window.EJS_alignStartButton;
     config.VirtualGamepadSettings = window.EJS_VirtualGamepadSettings;
-    config.buttonOpts = window.EJS_Buttons;
+
+    // set default button options
+    config.buttonOpts = {
```
ethanaobrien: I don't think these *should* be defined here. loader.js is meant to be a bare-bones loader, and if someone chooses to directly load emulator.js thats a lot of code for them to ship to get basic functions :)

ethanaobrien (dismissed):
Please fix the formatting before I continue reviewing; I don't have the patience at the moment to read through mixed changes

I do also think these separate buttons should be a separate `EJS_*` variable - this would keep formatting for each consistent (not needing to handle a boolean vs and object differently) and would completely remove the need for name-collision checks

michael-j-green:
> Please fix the formatting before I continue reviewing; I don't have the patience at the moment to read through mixed changes

No problem. I didn't notice the changes until after I had submitted the PR, and figured I would await your feedback on the button logic and implementation before fixing them.

I'll fix them and re-submit for review.

> I do also think these separate buttons should be a separate `EJS_*` variable - this would keep formatting for each consistent (not needing to handle a boolean vs and object differently) and would completely remove the need for name-collision checks

Do you mean leaving the existing EJS_Buttons behaviour as is (true/false controlling button visibility), and implementing a new EJS_ variable (for example EJS_ButtonPlayPause) to implement the different icon and custom callback functionality?

I had implemented the new functionality using the existing EJS_Buttons variable as it seemed like a logical place to configure it, and built it to accept either a boolean or array to maintain backwards compatibility.

The way the code is implemented in this PR is that if an unknown button name is supplied, nothing will change - it's simply ignored. So collisions cannot happen.

data/loader.js
michael-j-green: Will do. Give me a day or so to fix it and I'll resubmit.

data/loader.js
michael-j-green: Thanks! This is the kind of feedback I was fishing for :) I'll move it into the `emulator.js` file.

ethanaobrien:
>The way the code is implemented in this PR is that if an unknown button name is supplied, nothing will change - it's simply ignored. So collisions cannot happen.

Ah, that makes sense. I guess clean it up and I'll re-review

data/loader.js
michael-j-green: All formatting fixed.

Auto-formatting can be disabled in VSCode (for this repo only) by adding a file called `.vscode/settings.json`, and adding the following lines:
```json
{
    "diffEditor.ignoreTrimWhitespace": false,
    "editor.formatOnPaste": false,
    "editor.formatOnSave": false,
    "editor.formatOnSaveMode": "modifications"
}
```

Thinking we should add the above guidance somewhere to ensure the auto-formatting is disabled.

data/loader.js
michael-j-green: Code migrated.

data/loader.js
```diff
@@ -91,7 +91,7 @@
     config.adSize = window.EJS_AdSize;
     config.alignStartButton = window.EJS_alignStartButton;
     config.VirtualGamepadSettings = window.EJS_VirtualGamepadSettings;
-    config.buttonOpts = window.EJS_Buttons;
+    config.buttonUserOpts = window.EJS_Buttons;
```
michael-j-green: Host button configuration now applies to `config.buttonUserOpts` instead of `config.buttonOpts` as the later is used throughout the code and require a lot of refactoring to correct.

Buttons are set up beginning in emulator.js starting on line 210.

data/src/emulator.js
```diff
@@ -207,6 +207,8 @@ class EmulatorJS {
         if (this.debug || (window.location && ['localhost', '127.0.0.1'].includes(location.hostname))) this.checkForUpdates();
         this.netplayEnabled = (window.EJS_DEBUG_XX === true) && (window.EJS_EXPERIMENTAL_NETPLAY === true);
         this.config = config;
+        this.config.buttonOpts = this.defaultButtonOptions;
```
michael-j-green: `this.config.buttonOpts` is the variable that all toolbar buttons are built from. Here we set it to match the `this.defaultButtonOptions` which contains all of the default toolbar button settings. This is defined starting on line 1173.

data/src/emulator.js
```diff
@@ -1168,6 +1170,213 @@ class EmulatorJS {
...
+            callback: undefined
+        },
+        customBtn3: {
+            visible: false,
+            icon: null,
+            displayName: undefined,
+            callback: undefined
+        }
+    };
+    buildButtonOptions() {
```
michael-j-green: This function merges the settings provided by the host page via `this.config.buttonUserOpts` into `this.config.buttonOpts`.

index.html
```diff
@@ -245,6 +245,27 @@ <h1>EmulatorJS Demo</h1>
                 window.EJS_DEBUG_XX = enableDebug;
                 window.EJS_disableDatabases = true;
                 window.EJS_threads = enableThreads;
+                window.EJS_Buttons = {
```
michael-j-green: This is a sample usage of the `EJS_Buttons` variable, showing both existing usage (hiding buttons like restart by setting `restart: false`), and the new usage on `exitEmulation` and `customBtn`.

Custom icons are set by providing the svg XML. It is possible to set it to use an external file, but that might cause issues with paths when EJS is used in an iframe.

**Note**: this entire section will be removed before the PR is merged.

michael-j-green:
@ethanaobrien; I've cleaned and updated the code. It should be much easier to follow now :)

I've also inserted comments in the PR to hopefully explain how it all works.

data/loader.js
ethanaobrien: This would also break compatibility with anyone loading emulator.js directly, is there a reason for the change?

data/src/emulator.js
ethanaobrien: Shouldn't the `buildButtonOptions` function process this?

data/src/emulator.js
```diff
@@ -1168,6 +1170,213 @@ class EmulatorJS {
...
+        },
+        loadState: {
+            visible: true,
+            icon: '<svg viewBox="0 0 576 512"><path fill="currentColor" d="M572.694 292.093L500.27 416.248A63.997 63.997 0 0 1 444.989 448H45.025c-18.523 0-30.064-20.093-20.731-36.093l72.424-124.155A64 64 0 0 1 152 256h399.964c18.523 0 30.064 20.093 20.73 36.093zM152 224h328v-48c0-26.51-21.49-48-48-48H272l-64-64H48C21.49 64 0 85.49 0 112v278.046l69.077-118.418C86.214 242.25 117.989 224 152 224z"/></svg>',
+            displayName: "Load State",
+            callback: undefined
+        },
+        screenRecord: {
+            visible: true
+        },
```
ethanaobrien, line 1235: Should we also add a context menu option?

data/src/emulator.js
```diff
@@ -1168,6 +1170,213 @@ class EmulatorJS {
             elem.appendChild(elm);
         }
     }
+    defaultButtonOptions = {
+        playPause: {
+            visible: true,
+            icon: "play",
+            displayName: "Play/Pause",
+            callback: undefined
```
ethanaobrien: Does callback manually need to be set to undefined?

data/loader.js
michael-j-green: D'oh! You're right... I was doing it like this to keep the user supplied button options separate from the defaults and the active values.

I'll refactor clean this up.

data/src/emulator.js
michael-j-green: You're right, this is a better way. Just call the function from the get-go.

data/src/emulator.js
michael-j-green: We can indeed.

Though given previous discussions we've had, I had excluded it to prevent customisation of it. Do you want to allow hiding it? Changing the icon? Triggering a custom callback?

data/src/emulator.js
michael-j-green: No... that's not required. Because of the number of buttons I had to define, I just built a template and copy pasted it 20 times.

I'll remove the callback attribute from the defaults (it'll be undefined at that point anyway).

data/src/emulator.js
michael-j-green: Actually contextMenu is already in the list :) See line 1288

data/src/emulator.js
```diff
@@ -1379,10 +1577,10 @@ class EmulatorJS {
         });
         
         if (this.config.buttonOpts) {
-            if (this.config.buttonOpts.screenshot === false) screenshot.setAttribute("hidden", "");
-            if (this.config.buttonOpts.screenRecord === false) startScreenRecording.setAttribute("hidden", "");
-            if (this.config.buttonOpts.quickSave === false) qSave.setAttribute("hidden", "");
-            if (this.config.buttonOpts.quickLoad === false) qLoad.setAttribute("hidden", "");
+            if (this.config.buttonOpts.screenshot.visible === false) screenshot.setAttribute("hidden", "");
```
ethanaobrien, line 1563: These arent referenced anywhere else, right?

ethanaobrien:
I'm worried the internal API change may break some external users plugins (since apparently people use internal apis)... Would it be possible to instead of overwriting `this.config.buttonOpts` to create a seperate variable somewhere else? Or at that point this new option would need to be a seperate ejs option... What do you think?

data/src/emulator.js
michael-j-green: No... these entries are referencing the menu items under the context menu.

I do think there is value in duplicating these options to make them available in the toolbar (we can default the toolbar button visibility to hidden to maintain current behaviour) but for the purposes of this PR, lets try to maintain current behaviour as much as possible. Another PR could implement this.

michael-j-green:
> I'm worried the internal API change may break some external users plugins (since apparently people use internal apis)... Would it be possible to instead of overwriting `this.config.buttonOpts` to create a seperate variable somewhere else? Or at that point this new option would need to be a seperate ejs option... What do you think?

I understand this. My point of view though is that third parties should not be using internal API's. If third parties rely on an internal API, they should raise a PR to expose that API publicly.

The reason why I think this is that limiting ourselves to not modifying internal undocumented behaviour can restrict what changes/improvements/bug fixes/features we want to implement in the future.

Happy to plonk a "Possibly Breaking Change" tag on the PR with a note that if you're using an internal API related to buttons that you may need to change how you access it.

michael-j-green:
@ethanaobrien; if you approve this change for merging, I'll submit a PR containing supporting documentation for the website documentation, and remove the sample code index.html (which will revert it back to the current branch).

michael-j-green:
@ethanaobrien; any further changes you need made to this PR? :)

ethanaobrien:
Part of me feels like this needs to be a separate option - because while its a similar area its a completely different functionality and can cause over-complification of some sort if you disable most buttons and want to just add one

Though, it's just a thought. If youre ready this PR should be good to go

michael-j-green:
I do worry about the proliferation of EJS_ options which can make it tricky for devs to work with the project. I wouldn’t want to start creating a new variable for every button and every possible attribute we support as it adds complexity and a lot of technical debt we really don’t want :D

I’ve got just one final change to make - and that’s around the custom buttons. Currently the PR only supports three custom buttons. This is because the code as is just updates existing buttons defined in the default buttons list.

The updated code will look for any buttons defined in EJS_Buttons that have a displayName, icon, and callback and add them if they don’t match an existing button. Should have that ready reasonably soon.

ethanaobrien:
>I do worry about the proliferation of EJS_ options which can make it tricky for devs to work with the project. I wouldn’t want to start creating a new variable for every button and every possible attribute we support as it adds complexity and a lot of technical debt we really don’t want :D

Which part about having one option to disable buttons and one to add additional custom buttons makes it harder?

One part I worry about is the docs - having separate sections for separate versions adds complexity in my mind. Though I do realize the docs should state the version an option was added, stating "you can do this, or this if past a certain version" adds a bit of "Which way is better? Why was this changed? What version am I on?"

It should also allow us to do this properly as an array of buttons to add and wouldn't need to worry about accidentally attempting to overwrite an internal name or trying to distinguish between a setting for an internal button or a setting to add a separate button.

In the end, its whatever works. I don't think most will use the feature so it should be fine, though what're @allancoding's thoughts on this too?

michael-j-green:
I've cleaned up and implemented the rest of the requested changes. This PR can now handle unlimited custom buttons.

> Which part about having one option to disable buttons and one to add additional custom buttons makes it harder?

The current implementation for custom events is implemented using separate variables (EJS_onLoadState, etc), so I'm just thinking about those being created for each button which can get a bit messy adding support for extra buttons (especially custom ones).
 
> One part I worry about is the docs - having separate sections for separate versions adds complexity in my mind. Though I do realize the docs should state the version an option was added, stating "you can do this, or this if past a certain version" adds a bit of "Which way is better? Why was this changed? What version am I on?"

I think this is going to be an ongoing issue as new options are added over the project lifetime. We'll have to find a nice clean way to call this out.

I've noticed in some documentation, that they do essentially a separate set of documentation for each release version, so this is one possible solution - though it does increase the size of the project.

ethanaobrien:
>The current implementation for custom events is implemented using separate variables (EJS_onLoadState, etc), so I'm just thinking about those being created for each button which can get a bit messy adding support for extra buttons (especially custom ones).

Ah.. This is correct. I wasn't the one who implemented these so I didn't think about these - or I believe these may have been from when the project first started and was kept for backwards-compatibility. What should have been done (4.0+) is an on init hook, a callback to setup other callbacks once `window.EJS_emulator` was exposed - then the [`window.EJS_emulator.on`](https://github.com/EmulatorJS/EmulatorJS/blob/443a6f81cd182a010b353540d33b3a63f917afac/data/src/emulator.js#L382) function should be called. I've never introduced a fully breaking change when using as intended and I plan to keep it that way (even the 4.0 rewrite had no breaking changes like I originally thought it would)

>I think this is going to be an ongoing issue as new options are added over the project lifetime. We'll have to find a nice clean way to call this out.

Yes and no, in my mind its not much about the amount of options, its the choice between "if on this version, use this, otherwise use this"

>I've noticed in some documentation, that they do essentially a separate set of documentation for each release version, so this is one possible solution - though it does increase the size of the project.

This does greatly increase effort yeah - especially on versions not technically supported. simple notes added with the added version should be enough for this project

michael-j-green:
> Ah.. This is correct. I wasn't the one who implemented these so I didn't think about these - or I believe these may have been from when the project first started and was kept for backwards-compatibility. What should have been done (4.0+) is an on init hook, a callback to setup other callbacks once `window.EJS_emulator` was exposed - then the [`window.EJS_emulator.on`](https://github.com/EmulatorJS/EmulatorJS/blob/443a6f81cd182a010b353540d33b3a63f917afac/data/src/emulator.js#L382) function should be called. I've never introduced a fully breaking change when using as intended and I plan to keep it that way (even the 4.0 rewrite had no breaking changes like I originally thought it would)

Yup... and non-breaking was my goal for this PR too, which is why this change accepts EJS_Buttons as it is now, while also accepting the new format.

Breaking changes suck :D And that's why I'm putting in so much discussion on this. I don't want to break things, and I also don't want to set an inappropriate precedent.

> > I think this is going to be an ongoing issue as new options are added over the project lifetime. We'll have to find a nice clean way to call this out.
> 
> Yes and no, in my mind its not much about the amount of options, its the choice between "if on this version, use this, otherwise use this"
> 
> > I've noticed in some documentation, that they do essentially a separate set of documentation for each release version, so this is one possible solution - though it does increase the size of the project.
> 
> This does greatly increase effort yeah - especially on versions not technically supported. simple notes added with the added version should be enough for this project

Sweet :) I included version numbers in the documentation for this change so that users are well informed.

index.html
allancoding: Yeah just remove this and it will look good.

allancoding (changes requested):
Remove the example :)

I do agree that not very many people are going to use this. But in my opinion it really cleans it up will all the svg icons.

About documenting the options and what versions the came out in... I think it would be nice but not necessary. They could go back in the release and look for themselves. Staying on older versions should not be recommended, most people will update because they want the new features and or better performance.

index.html
michael-j-green: Done!

PR is ready for merging.

allancoding (approved):

ethanaobrien (approved):
I don't entirely agree with the implementation, but I guess if it works it works

PR #1001, opened 2025-04-30
Virtual Gamepad Fix

allancoding:
I found this to be the best way to decide whether to show the Menu Bar Button.

Many devices will have a number bigger than `0` for the `navigator.maxTouchPoints` value so...
If the device is mobile then show the button by default.
If the device has touch enabled and they do not have a mouse then enable.
If the device has touch but has a mouse then the button can be hidden.

To include  support for devices that have a mouse and touch screen the button can be turned on in settings.
Since they have a mouse they should be able to enable it. 
However when disabling it in the settings it only makes the button invisible and will still work.


data/src/emulator.js
```diff
@@ -236,6 +236,21 @@ class EmulatorJS {
...
             return check;
         })();
+        this.hasTouchScreen = (function() {
+            var result = false;
+            if (window.PointerEvent && ('maxTouchPoints' in navigator)) {
+                if (navigator.maxTouchPoints > 0) {
+                    result = true;
```
ethanaobrien: literally just return here lol, then return false at the end

data/src/emulator.js
```diff
@@ -1368,7 +1383,7 @@ class EmulatorJS {
         }
         this.addEventListener(this.elements.contextmenu, 'contextmenu', (e) => e.preventDefault());
         this.addEventListener(this.elements.parent, 'contextmenu', (e) => e.preventDefault());
-        this.addEventListener(this.game, 'mousedown', hideMenu);
+        this.addEventListener(this.game, 'mousedown touchend', hideMenu);
```
ethanaobrien, line 1389: What is the intention here

data/src/emulator.js
```diff
@@ -3956,12 +3971,16 @@ class EmulatorJS {
...
+        if (this.touch || this.hasTouchScreen) {
             const menuButton = this.createElement("div");
             menuButton.innerHTML = '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path d="M0 96C0 78.33 14.33 64 32 64H416C433.7 64 448 78.33 448 96C448 113.7 433.7 128 416 128H32C14.33 128 0 113.7 0 96zM0 256C0 238.3 14.33 224 32 224H416C433.7 224 448 238.3 448 256C448 273.7 433.7 288 416 288H32C14.33 288 0 273.7 0 256zM416 448H32C14.33 448 0 433.7 0 416C0 398.3 14.33 384 32 384H416C433.7 384 448 398.3 448 416C448 433.7 433.7 448 416 448z"/></svg>';
             menuButton.classList.add("ejs_virtualGamepad_open");
             menuButton.style.display = "none";
-            this.on("start", () => menuButton.style.display = "");
+            this.on("start", () => {
+                if (this.isMobile || !matchMedia('(pointer:fine)').matches) {
+                    menuButton.style.display = ""
+                }
```
ethanaobrien: This should always be equal to `this.hasTouchScreen`. Never will we 100% know there is no mouse depending on the type of touchscreen or any additional accessories attached to the device

You can always disable the virtual gamepad, but without this button you can *never* enable it. This is a "better safe than sorry" situation.

data/emulator.css
```diff
@@ -1337,14 +1337,15 @@
 
 .ejs_virtualGamepad_open {
     display: inline-block;
-    width: 30px;
-    height: 30px;
+    width: 1.5em;
```
ethanaobrien: Why do we want this relative to font size?

ethanaobrien (changes requested):
I will need to run tests on additional devices.

This is a big change that if done incorrectly could cause a large amount of mobile users unusable

data/src/emulator.js
allancoding: This makes so that you can click off the popup to close the context menu on mobile.

data/emulator.css
allancoding: It will scale better across devices.

ethanaobrien: The toolbar menu at that bottom does not scale to font size, this should not either

data/src/emulator.js
allancoding: Ok I set to be invisible if there is a mouse.

data/src/emulator.js
```diff
@@ -4729,12 +4751,16 @@ class EmulatorJS {
...
-        if (this.touch || navigator.maxTouchPoints > 0) {
+        if (this.touch || this.hasTouchScreen) {
             const virtualGamepad = createSettingParent(true, "Virtual Gamepad", home);
             addToMenu(this.localization('Virtual Gamepad'), 'virtual-gamepad', {
                 'enabled': this.localization("Enabled"),
                 'disabled': this.localization("Disabled")
             }, this.isMobile ? 'enabled' : 'disabled', virtualGamepad, true);
+            addToMenu(this.localization('Menu Bar Button'), 'menu-bar-button', {
+                'show': this.localization("show"),
+                'hide': this.localization("hide")
```
ethanaobrien: these should probably be "visible" and "hidden"

data/src/emulator.js
```diff
@@ -4729,12 +4751,16 @@ class EmulatorJS {
...
+            }, (this.isMobile || !matchMedia('(pointer:fine)').matches) ? 'show' : 'hide', virtualGamepad, true);
```
ethanaobrien: This should default to show. Logic for hiding on start is done separately

ethanaobrien (changes requested):

data/src/emulator.js
allancoding: But then the button may be hidden but the setting will say its visible.

ethanaobrien: Then set the default value here instead of above

ethanaobrien (commented):
@michael-j-green and @S-Coding23 Please test this PR on every device y'all have

ethanaobrien (approved):
Should probably be fine

Hopefully nothing breaks!

Please review and merge @michael-j-green  or @S-Coding23

michael-j-green:
Going to try and get this tested and reviewed this weekend.

michael-j-green:
Can confirm this works properly on macOS Safari and Edge (screenshot of Safari only as Edge is the same):
Screenshot 2025-05-25 at 12 19 15 pm

Here's iOS using the Xcode iPhone Simulator:
Screenshot 2025-05-25 at 12 26 57 pm

And iPadOS using the Xcode iPad Simulator:
Screenshot 2025-05-25 at 12 30 08 pm

Let me know if you need anything else, and apologies for the delays.

allancoding:
Looks like it works.

michael-j-green (approved):
Approved

PR #1004, opened 2025-05-01
Add demo autoload

allancoding:
The folder `roms/` is in the `.gitignore` file.
Using the `rom` _urlParam_ you can set a file name and it will autoload a rom from the `roms/` folder.

Also added padding and centered text on input box.

### DIFF
--- a/index.html
+++ b/index.html
@@ -40,6 +40,8 @@
                 font-weight: bold;
                 font-size: 20px;
                 margin: 5px;
+                text-align: center;
+                padding: 10px;
             }
 
             #box:hover, #box[drag] {
@@ -54,7 +56,7 @@
                 top: 0;
                 width: 100%;
                 height: 100%;
-                opacity: 0
+                opacity: 0;
             }
 
             #display {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -110,14 +113,14 @@ <h1>EmulatorJS Demo</h1>
             let enableThreads = false;
             const queryString = window.location.search;
             const urlParams = new URLSearchParams(queryString);
-            if (urlParams.get('debug') == 1) {
+            if (urlParams.get('debug') == 1 || urlParams.get('debug') == "true") {
```
ethanaobrien: Please use `===` (yes, this appears to have been incorrect before too.. or actually maybe it wasn't. Best to also parseint on that `.get("debug")`)

index.html
```diff
@@ -110,14 +113,14 @@ <h1>EmulatorJS Demo</h1>
...
                 enableDebug = true;
                 console.log("Debug is enabled");
             } else {
                 console.log("Debug is disabled");
             }
 
-            if (urlParams.get('threads') == 1) {
+            if (urlParams.get('threads') == 1 || urlParams.get('threads') == "true") {
```
ethanaobrien: here too

ethanaobrien (changes requested):

index.html
michael-j-green: I think this might have been me sometime last year, as I added the urlParams for `debug` and `threads`.

ethanaobrien (approved):

michael-j-green (approved):
Approved

PR #1005, opened 2025-05-01
Fix Error Screen

allancoding:
- Removes box shadow on the error screen
- Hides virtual gamepad on error screen


ethanaobrien:
Please screenshot how this looks

allancoding:
Old:
image
image
Fix:
image
image

ethanaobrien:
IMO the black/faded background on the menu bar should still be there. Helps make it easier to see, and that something actually showed up

allancoding:
You still see it when the setting button appears.

michael-j-green:
> You still see it when the setting button appears.

That's not reflected in the screenshot ;)

michael-j-green:
Just wondering if this PR can be extended to implement something we talked about late last year; which was for more helpful error messages to be displayed instead of simply "Network Error", or "Failed to start game".

Would be good to give the user more feedback as to what went wrong.

I think the idea we had was for a generic error to be displayed, with a dropdown that when clicked/tapped displayed a more detailed error.

allancoding:
> Just wondering if this PR can be extended to implement something we talked about late last year; which was for more helpful error messages to be displayed instead of simply "Network Error", or "Failed to start game".
> 
> Would be good to give the user more feedback as to what went wrong.
> 
> I think the idea we had was for a generic error to be displayed, with a dropdown that when clicked/tapped displayed a more detailed error.

Yeah that would be useful

ethanaobrien:
>Would be good to give the user more feedback as to what went wrong.

A majority of users that can make sense of this information will be able to use the console, similar to the "error for site owner, check console", all errors are logged to the console and dumping them to the UI can lead to confusion for the 80% (its probably more) that won't be able to do a thing with this information

I don't think its needed

michael-j-green:
I think that’s mostly true. I think there’s room for more descriptive errors that aren’t full debug messages.

I’m thinking more “lightweight” errors for more casual users who aren’t technical enough to open the console.

Things like “Unable to extract BIOS archive”, “Unable to extract ROM archive”, “Unable to download requested core” (in the case of the core being fetched from the CDN), etc.

Nothing too detailed, but just enough to be helpful.

allancoding:
image
image

data/src/emulator.js
```diff
@@ -599,7 +599,7 @@ class EmulatorJS {
                     if (!this.supportsWebgl2) {
                         this.startGameError(this.localization('Outdated graphics driver'));
                     } else {
-                        this.startGameError(this.localization('Network Error'));
+                        this.startGameError(this.localization('Network Error') + ": " + this.localization("While Downloading Core") + "(" + filename + ")");
```
ethanaobrien: "Error downloading core"

No need to add filename. Thats logged to the console.

data/src/emulator.js
```diff
@@ -663,7 +663,7 @@ class EmulatorJS {
                 this.textElem.innerText = this.localization("Download Game State") + progress;
             }, true, { responseType: "arraybuffer", method: "GET" }).then((res) => {
                 if (res === -1) {
-                    this.startGameError(this.localization('Network Error'));
+                    this.startGameError(this.localization('Network Error') + ":" + this.localization("When Downloading Game State"));
```
ethanaobrien: "Error downloading game state"

data/src/emulator.js
```diff
@@ -711,7 +711,7 @@ class EmulatorJS {
                 this.textElem.innerText = progressMessage + progress;
             }, true, { responseType: "arraybuffer", method: "GET" });
             if (res === -1) {
-                this.startGameError(this.localization("Network Error"));
+                this.startGameError(this.localization('Network Error')+"\n"+this.localization("Check console"));
```
ethanaobrien: Please revert/change to using double quotes, and I don't think we should specify "Check console" here.

data/src/emulator.js
```diff
@@ -856,7 +856,7 @@ class EmulatorJS {
                     this.textElem.innerText = this.localization("Download Game Data") + progress;
                 }, true, { responseType: "arraybuffer", method: "GET" });
                 if (res === -1) {
-                    this.startGameError(this.localization('Network Error'));
+                    this.startGameError(this.localization('Network Error')+"\n"+this.localization("Check console"));
```
ethanaobrien: same here

data/src/emulator.js
```diff
@@ -909,7 +909,7 @@ class EmulatorJS {
     initModule(wasmData, threadData) {
         if (typeof window.EJS_Runtime !== "function") {
             console.warn("EJS_Runtime is not defined!");
-            this.startGameError(this.localization("Failed to start game"));
+            this.startGameError(this.localization("Failed to start game")+"\n"+this.localization("Runtime not defined"));
```
ethanaobrien: "Error loading EmulatorJS runtime"

ethanaobrien (changes requested):
I really dont think we need the additional "check console" messages, or it should be appended in the `startGameError` function when debug is enabled (though at this point, the dev **really, really *should* know what they're doing**) , and idk but the new css changes reeeeeaaaalllly stands out.

ethanaobrien:
@michael-j-green Please also leave your review

data/src/emulator.js
allancoding: I feel like this is important. The filename tells what core failed to download.

ethanaobrien: The filename is logged in the console. 95% of users cant do anything with this information

data/src/emulator.js
allancoding: I feel like 'Network Error' is very vague

ethanaobrien: "Check console" doesnt add any details

allancoding:
> I really dont think we need the additional "check console" messages, or it should be appended in the `startGameError` function when debug is enabled (though at this point, the dev **really, really _should_ know what they're doing**) , and idk but the new css changes reeeeeaaaalllly stands out.

Well you did say that it made it harder to see...

> IMO the black/faded background on the menu bar should still be there. Helps make it easier to see, and that something actually showed up

data/src/emulator.js
allancoding: I actually feel like it does. If a person just sees `Network Error` they may not know what to do, like at all. Including the `Check console` gives them a next step to fix the issue.

ethanaobrien: You've seen people in issues and the discord when we ask for them to check the console, right?

allancoding: Yes I have 😄
Ok 1st in my option giving them a error `Network Error` they are going to come straight to EmulatorJS for questions.
2nd if they see `Check console` and don't know what it means chances are they will ask google what a console log is then they are on the road to solve the error for themselves.
3rd if they come to EmulatorJS with the error in the console it's one less step we have to give telling them to check the console.

data/src/emulator.js
michael-j-green: I agree with dropping the "Check Console" part. A dev is going to know to do that anyway, and for the average user opening the console (especially on a mobile device) can be non-trivial.

At most we should have a drop down/more details link that expands on the error. In the case of "Error downloading core", this could be because the CDN is uncontactable, a 404 error, or some other condition which is something that's useful to convey to the end user.

data/src/emulator.js
michael-j-green: > You've seen people in issues and the discord when we ask for them to check the console, right?

If I may point out, a Gaseous or Romm user won't be going to EJS (Discord, Matrix, website, etc) for details. As far as they're concerned EJS is a part of Gaseous. It's akin to thinking someone is going to go to Tata because they're having an issue with the airbag in their Ford.

And if they do try to go to EmulatorJS - they may end up in the wrong project.

ethanaobrien:
I personally think the css changes should be reverted

allancoding:
I like the look of nothing better. On desktop I could expand the settings icon to have the `Settings` text next to it.

allancoding:
I like text shadow better:
image
image
I also fix the color on mobile layout, it was inheriting the wrong color:
image

michael-j-green:
Thinking we need a background on the text. Red on a dark colour has too little contrast rendering it a bit harder to read.

Perhaps a red box (light red, with a contrast red border might be best), with white error message text on top?

allancoding:
I come back to:
image
image
Just adds shadow.

ethanaobrien:
Could you screenshot a larger area of the screen

allancoding:
image
image

ethanaobrien:
This is fine. What does @michael-j-green think?

michael-j-green:
Looks alright to me... would still like to see a contrasting background behind the red text.

allancoding:
Ok I'll add that.

allancoding:
Small text shadow on loading text:
image

ethanaobrien:
@allancoding 

Devtools -> (Possibly "Add to activity bar", the plus button on the right) -> CSS overview -> "Capture Overview" -> "Contrast Issues"

allancoding:
Match the style:
image

allancoding:
@michael-j-green Is that what you were looking for?

michael-j-green:
You'll have to excuse the crudity of this example - it's only meant to demonstrate possible colours and contrast, but I was thinking something more along the lines of this:
Screenshot 2025-06-20 at 10 12 00 am
(Imagine a rounded border with drop shadow and a serif font :D )

Having a good contrast is an important accessibility feature that users who might have vision impairment would appreciate.

ethanaobrien:
I think the latest example Allan sent is fine - I think the text just needs to be bigger and bolded, with a larger element margin

michael-j-green:
How about bringing up the brightness of the red text a bit?
Screenshot 2025-06-20 at 11 16 07 am

allancoding:
How's this:
image

michael-j-green:
Yup. That's far better.

michael-j-green (approved):
Appproved

ethanaobrien (approved):

PR #1009, opened 2025-05-09
Mouse Fix & Controls CSS

allancoding:
Added hover for other players:
image
Fix alignment and size:
image
Before:
image

Changes the mouse movement listener.
### Before:
- Any movement  shows menu bar.

### Now:
- Any down movement in a 12 deg angle
- Movement in the zone of the menu bar.


ethanaobrien:
I really don't think we should change the default cursor action. This adds a loooot of additional computation run several times per second for 1% of users

allancoding:
I actually think it improves performance by not running a new setTimeout on every mouse movement:
```
clearTimeout(timeout);
timeout = setTimeout(hide, 3000);
```

ethanaobrien:
I dont think it should be changed by default.

ethanaobrien (changes requested):
Please add an option to switch between the two behaviours

allancoding:
Added setting option:
image
This menu can be expanded later to include pointer lock settings.

ethanaobrien (approved):
@michael-j-green Please review this PR

michael-j-green:
Tested on macOS Safari. Toolbar displays when moving downwards in diagonally more than a 12 degree angle, and when in the toolbar region.

Settings option to revert to original behaviour works.

michael-j-green (approved):
Approved

PR #1038, opened 2025-07-02
Localization and logging fix

incredibleIdea:
- Add .gitattributes to enforce consistent line endings (EOL)
- Fix console logging when text length is 0
- Improve console message context
- Update zh-CN localization


ethanaobrien (approved):
Looks good, thank you.

Pending approval by another contributor

michael-j-green (approved):
Happy to accept this, but would prefer to see unrelated changes submitted as separate PR's.

In this case, logging, localisation, and console output would/should all be separate PR's.

Regardless, approved, and thankyou! :)

PR #1045, opened 2025-07-08
Fix for context menu

### LINKED ISSUE #1041: Button issues
Image

Image

michael-j-green:
This change implements the following:

With the following:
```javascript
EJS_Buttons = {
   contextMenu: false
}
```

The following should happen:
* `isMobile === true`: the context menu should be displayed
* `isMobile === false`: the context menu should be hidden

This is intended behaviour that was reverted due to changes in the button options.

Also to assist with debugging, this change also implements a new variable named `EJS_browserMode` which can be used to force desktop or browser mode.

`EJS_browserMode` will accept the following values:
- `0` (default) = automatic
- `1` or `mobile` = mobile mode
- `2` or `desktop` = desktop mode

When accessing the `index.html` page included, appending `?browserMode=<mode>` to the URL will force EJS to run in the specified browser mode.


data/loader.js
```diff
@@ -123,6 +123,9 @@
     config.videoRotation = window.EJS_videoRotation;
     config.hideSettings = window.EJS_hideSettings;
     config.shaders = Object.assign({}, window.EJS_SHADERS, window.EJS_shaders ? window.EJS_shaders : {});
+    if (window.EJS_forceMobile === true) {
```
michael-j-green: This was added for debugging purposes and seemed like a useful addition.

However, I'm more than happy to remove it as I've added it without discussion.

data/src/emulator.js
```diff
@@ -230,9 +230,12 @@ class EmulatorJS {
             this.config.adSize = (Array.isArray(this.config.adSize)) ? this.config.adSize : ["300px", "250px"];
             this.setupAds(this.config.adUrl, this.config.adSize[0], this.config.adSize[1]);
         }
-        this.isMobile = (function() {
+        this.isMobile = (() => {
```
michael-j-green, line 249: This (and the changing of the function to a lambda above) support the forceMobile option.

index.html
```diff
@@ -132,6 +133,13 @@ <h1>EmulatorJS Demo</h1>
                 console.log("Threads are disabled");
             }
 
+            if (parseInt(urlParams.get('forceMobile')) === 1 || urlParams.get('forceMobile') === "true") {
```
michael-j-green: This is part of the forceMobile option. Add `?forceMobile=1` to the URL of the index.html page, and it will force EJS to load as if it had detected it was running on mobile.

data/loader.js
allancoding: I think it is useful, I would change it so that if it is `false` then it forces desktop.

data/src/emulator.js
```diff
@@ -2288,7 +2286,7 @@ class EmulatorJS {
                 pauseButton.style.display = "none";
                 playButton.style.display = "none";
             }
-            if (this.config.buttonOpts.contextMenuButton === false && this.config.buttonOpts.rightClick !== false && this.isMobile === false) contextMenuButton.style.display = "none"
+            if (this.config.buttonOpts.contextMenu.visible === false && this.config.buttonOpts.rightClick !== false && this.isMobile === false) contextMenuButton.style.display = "none"
```
ethanaobrien, line 2314: Oh, this was the only issue?

michael-j-green: Well, that and line 1344 where when `contextMenu` was used as a button key it aborted early.

data/loader.js
michael-j-green: Not sure I like that, because it makes the logic a bit whacky.

How about a number?
- 0 (default) = do nothing
- 1 = force mobile
- 2 = force desktop

@ethanaobrien; what do you think?

allancoding: Works for me

michael-j-green: Might rename the variable from `forceMobile` to something like `browserMode`, because then the possible values make sense:
- `0` (default) = automatic
- `1` or `mobile` = mobile mode
- `2` or `desktop` = desktop mode

data/loader.js
```diff
@@ -123,6 +123,24 @@
     config.videoRotation = window.EJS_videoRotation;
     config.hideSettings = window.EJS_hideSettings;
     config.shaders = Object.assign({}, window.EJS_SHADERS, window.EJS_shaders ? window.EJS_shaders : {});
+    if (window.EJS_browserMode) {
```
michael-j-green: Accepts a number as either a number or a string and a string to configure the browser mode

index.html
michael-j-green: Changed to `browserMode`. See PR description.

data/src/emulator.js
```diff
@@ -230,7 +230,15 @@ class EmulatorJS {
             this.config.adSize = (Array.isArray(this.config.adSize)) ? this.config.adSize : ["300px", "250px"];
             this.setupAds(this.config.adUrl, this.config.adSize[0], this.config.adSize[1]);
         }
-        this.isMobile = (function() {
+        this.isMobile = (() => {
+            // browserMode can be either a 1 (force mobile), 2 (force desktop) or undefined (auto detect)
+            switch (this.config.browserMode) {
```
ethanaobrien, line 251: Honestly, I believe this would make more sense to be a string rather than an int, what do you think?

michael-j-green: I'm happy either way. The only reason this exists is because I needed to test both desktop and mobile versions easily and simply forgot to remove this option in the original commit - it then seemed like a good enough dev assist to leave in :D

This int is parsed starting at line 126 in `loader.js`. That will take an int (as a string or int) or a string such as `desktop` or `mobile` and stores the correct int to select the browser mode in `config.browserMode`.

If you want to leave the feature in, we can certainly move that logic into the `isMobile` function, or up in the EmulatorJS constructor (which now I think about it might be the best place for it).

However, in the `isMobile` function I like the idea that we're only checking for ints as the value is normalised, it keeps the code a bit cleaner as we don't have to evaluate a string, and if we want to use the `config.browserMode` value elsewhere, we only have to evaluate which int it is, and not repeat the string evaluation.

ethanaobrien: > This int is parsed starting at line 126 in loader.js.

> If you want to leave the feature in, we can certainly move that logic into the isMobile function, or up in the EmulatorJS constructor (which now I think about it might be the best place for it).

Ah, I do see now. I do think we do need to evaluate it at some point in `emulator.js`. `loader.js` should only be barebones for loading settings, not so much validation (because there *are* users that I've seen directly making their own `EmulatorJS` instance, rather than using loader). I believe the most complicated thing in `loader.js` is the loading a language file - and the only reason for this is there was no easy solution for running this asynchronous once setting up the class without a mess :)

michael-j-green: Already started moving it to the constructor 😉

Pretty much as soon as I hit send on my last comment I thought "yeah, it *should* move to the constructor", for exactly the reasons you pointed out (and it just makes more sense there anyway organisationally).

I'll post the update soon.

michael-j-green: `loader.js` has now been reverted to the original, and the `browserMode` logic has been moved to the `emulatorJS` constructor.

Feels a lot cleaner now.

michael-j-green:
@ethanaobrien; one last thing before this can be completed: as it stands, the context menu follows the logic as set out (set visibility, depending on browser mode).

However; it can be renamed, or a new icon assigned - is this a behaviour you're happy to keep, or did you want to block all changes to this menu (aside from the aforementioned visibility (depending on browser mode))?

allancoding (dismissed):
Yeah, keeping the loader.js minimal is better.

data/src/emulator.js
```diff
@@ -209,6 +209,22 @@ class EmulatorJS {
         this.config = config;
         this.config.buttonOpts = this.buildButtonOptions(this.config.buttonOpts);
         this.config.settingsLanguage = window.EJS_settingsLanguage || false;
+        switch (window.EJS_browserMode) {
```
ethanaobrien: This should be a this.config option set in loader.js, then loaded and checked here

index.html
```diff
@@ -132,6 +133,10 @@ <h1>EmulatorJS Demo</h1>
                 console.log("Threads are disabled");
             }
 
+            if (urlParams.get('browserMode')) {
+                browserMode = urlParams.get('browserMode');
+            }
```
ethanaobrien: We've also standardized using double quotes `"`, similar to many other languages that require it. I don't know if the index.html file has been updated though

data/src/emulator.js
michael-j-green: Agreed... though in the line immediately before this one, we're referencing `window.EJS_settingsLanguage`, and I believe there are other examples too.

I'll update this one to a `this.config` option, but I'm thinking one of us should do a clean up of other `window.EJS_*` variables that are set here to match so that we have consistency. What do you think?

index.html
michael-j-green: There were other entries in index.html that had single quotes. These have all now been updated to double quotes ".

data/src/emulator.js
michael-j-green: I just did a quick search. `EJS_` appears 226 times in `emulator.js`:
Screenshot 2025-07-11 at 11 52 54 am

So making this consistent is going to be fun...

michael-j-green: The latest commit removes the reference to `window.EJS_browserMode` from `emulator.js`.

ethanaobrien: >I just did a quick search. EJS_ appears 226 times in emulator.js:

A majority of these are likely constructors or classnames :)

>I'll update this one to a this.config option, but I'm thinking one of us should do a clean up of other window.EJS_* variables that are set here to match so that we have consistency. What do you think?

Sounds good!

index.html
ethanaobrien: Thank you

data/src/emulator.js
```diff
@@ -209,6 +209,22 @@ class EmulatorJS {
...
+            case 1: // Force mobile
+            case "1":
+            case "mobile":
+                console.log("Force mobile mode is enabled");
+                this.config.browserMode = 1;
+                break;
+            case 2: // Force desktop
+            case "2":
+            case "desktop":
+                console.log("Force desktop mode is enabled");
```
ethanaobrien: I'm sorry -- oone last thing - We shouldn't dump any logs unless `EJS_DEBUG_XX` is enabled (You'll see this if you search around the file)

data/src/emulator.js
michael-j-green: Lets make @allancoding do it :D

michael-j-green:
> @ethanaobrien; one last thing before this can be completed: as it stands, the context menu follows the logic as set out (set visibility, depending on browser mode).
> 
> However; it can be renamed, or a new icon assigned - is this a behaviour you're happy to keep, or did you want to block all changes to this menu (aside from the aforementioned visibility (depending on browser mode))?

@ethanaobrien; could you comment on the above? :)

ethanaobrien:
>However; it can be renamed, or a new icon assigned - is this a behaviour you're happy to keep, or did you want to block all changes to this menu (aside from the aforementioned visibility (depending on browser mode))?

Great question - I do agree with your idea, that's probably the best way to go about it

michael-j-green:
> > However; it can be renamed, or a new icon assigned - is this a behaviour you're happy to keep, or did you want to block all changes to this menu (aside from the aforementioned visibility (depending on browser mode))?
> 
> Great question - I do agree with your idea, that's probably the best way to go about it

Cool :) I'll block all changes except for the visibility of the icon. From now on, it will always have the burger icon, and the same name.

data/src/emulator.js
allancoding: Oh no looks like i've been voluntold.

michael-j-green:
> > > However; it can be renamed, or a new icon assigned - is this a behaviour you're happy to keep, or did you want to block all changes to this menu (aside from the aforementioned visibility (depending on browser mode))?
> > 
> > 
> > Great question - I do agree with your idea, that's probably the best way to go about it
> 
> Cool :) I'll block all changes except for the visibility of the icon. From now on, it will always have the burger icon, and the same name.

This change has now been committed.

ethanaobrien (approved):

allancoding (approved):

PR #1059, opened 2025-08-16
EJS_dontExtractBIOS doesn't work if BiosUrl has "/" in its path

daguile:
Game doesn't load and get stuck at Download BIOS step when bios url has / in its path and EJS_dontExtractBIOS set to true.
This was due to the fact that bios path was not handled the same way than when bios was extracted

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -729,14 +729,14 @@ class EmulatorJS {
                 return resolve(assetUrl);
             }
             const gotData = async (input) => {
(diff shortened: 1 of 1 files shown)


ethanaobrien (approved):
Thank you. Awaiting another admin approval

allancoding (approved):

daguile:
Made a stupid mistake while rewriting my PR. Will submit a new PR to fix it

PR #1064, opened 2025-08-19
Improve multi-disc handling for dosbox-pure

### LINKED ISSUE #1061: Can't play multi disc dos games
Hi, I'm trying to play some dos games with dosbox-pure core. My games are in zip file format with floppy disc images inside (.img). Everything works fine if only one image is in the zip file, but if the game has several floppy images then I'm stuck on the installation phase because I can't switch to the other floppy image. In standard RA I can  just use "Disc Control" option to to switch floppy but in EmulatorJS, I didn't succeed in bringing the RA menu to show (F1 does not work). Am I missing something ? Are there some workarounds ? Thanks in advance for your help

daguile:
dosbox-pure directly handles zip file and configures muti-disc support if several images are found in the zip. Therefore, we must not extract zip file when dealing with this case.

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -202,6 +202,7 @@ class EmulatorJS {
     constructor(element, config) {
         this.ejs_version = "4.3.0-beta";
         this.extensions = [];
+        this.allSettings = {};
         this.initControlVars();
         this.debug = (window.EJS_DEBUG_XX === true);
         if (this.debug || (window.location && ["localhost", "127.0.0.1"].includes(location.hostname))) this.checkForUpdates();
@@ -809,17 +810,19 @@ class EmulatorJS {
             this.textElem.innerText = this.localization("Download Game Data");
 
             const gotGameData = (data) => {
-                if (["arcade", "mame"].includes(this.getCore(true))) {
-                    this.fileName = this.getBaseFileName(true);
+                const altName = this.getBaseFileName(true);
+
+                let ext = altName.split(".").pop().toLowerCase();
+                const coreName = this.getCore(true);
+                if (["arcade", "mame"].includes(coreName) || (coreName === "dos" && ext === "zip")) {
+                    this.fileName = altName;
                     this.gameManager.FS.writeFile(this.fileName, new Uint8Array(data));
                     resolve();
                     return;
                 }
 
-                const altName = this.getBaseFileName(true);
-
                 let disableCue = false;
-                if (["pcsx_rearmed", "genesis_plus_gx", "picodrive", "mednafen_pce", "smsplus", "vice_x64", "vice_x64sc", "vice_x128", "vice_xvic", "vice_xplus4", "vice_xpet", "puae"].includes(this.getCore()) && this.config.disableCue === undefined) {
+                if (["pcsx_rearmed", "genesis_plus_gx", "picodrive", "mednafen_pce", "smsplus", "vice_x64", "vice_x64sc", "vice_x128", "vice_xvic", "vice_xplus4", "vice_xpet", "puae"].includes(coreName) && this.config.disableCue === undefined) {
                     disableCue = true;
                 } else {
                     disableCue = this.config.disableCue;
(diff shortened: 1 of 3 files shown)


michael-j-green:
Thanks for your submission.

Looking at the code changes, I have some reservations about functionality that is affected, but I'm putting my little one to bed right now so can't review properly.

Give me about two hours. I'll pull this branch and test it.

michael-j-green (dismissed):
Looks really good!

Spent too much time playing Doom

data/src/emulator.js
```diff
@@ -202,6 +202,7 @@ class EmulatorJS {
     constructor(element, config) {
         this.ejs_version = "4.3.0-beta";
         this.extensions = [];
+        this.allSettings = {};
```
ethanaobrien, line 205: Why is this added with no additional uses?

data/src/emulator.js
```diff
@@ -809,17 +810,19 @@ class EmulatorJS {
...
 
             const gotGameData = (data) => {
-                if (["arcade", "mame"].includes(this.getCore(true))) {
-                    this.fileName = this.getBaseFileName(true);
+                const altName = this.getBaseFileName(true);
+
+                let ext = altName.split(".").pop().toLowerCase();
+                const coreName = this.getCore(true);
+                if (["arcade", "mame"].includes(coreName) || (coreName === "dos" && ext === "zip")) {
+                    this.fileName = altName;
```
ethanaobrien: I don't think it should be done this way:

EmulatorJS supports multiple archive formats. It checks if the file is archived by the header bytes, not the extension (we do not always have this)
Specifically whitelisting the `zip` extension will break if you ever try to feed it something like a blob

I think we should simply revert this part of the code, and add "dos" to that initial arcade/mame check, if the goal is to never extract and let dosbox handle the file.

ethanaobrien (changes requested):

data/src/emulator.js
daguile: Wihout this, there is an exception is thrown in `setupDisksMenu` because `this.allSettings` is not initialized. But you made me realize that I forgot to remove the previous initialization in `setupSettingsMenu`

michael-j-green: @ethanaobrien; I'm getting an error that references this.allSettings with the Amiga core:
Screenshot 2025-08-20 at 4 27 21 pm

So I think there is a more generic error with multiple disk handling.

Perhaps @daguile's fix is enough to resolve that?

data/src/emulator.js
daguile: I see what you mean.
The only minor problem is that dosbox does not directly support all the archive formats like `7z` for example. I will check that.
Furthermore, dosbox supports more file formats like .com or `.exe` or `.bat`. In that case, we might want to wrap a bunch of files up in an archive file. So this is not an easy case.
Maybe the decision to extract or not an archive could be made downstream so that we can still handle blobs correctly.
Let me know what you prefer or suggest

data/src/emulator.js
michael-j-green: This is in the EmulatorJS constructor, and all three other "this.allSettings" entries I can find are also in this same class - none of which are guarded for null/undefined. This makes sense to at least initialise it here.

I've just tested it with the Amiga core with a multidisk game I was having issues with, and this fix resolves it.

data/src/emulator.js
ethanaobrien: I think one solution that may work is having the opt-in ability to not decompress the rom, similar to dontDecompressBios - this could also help for more than just dosbox for systems that support a certain archive format and the website owner would rather let the core handle it

daguile: Fair enough. I will make modifications in that sense

daguile: Done

ethanaobrien (approved):
Looks good, thanks!

allancoding (approved):
Tested + Works

michael-j-green:
This introduces a new option called `EJS_dontExtractRom`. I'll be happy to approve this once a PR for an update to the documentation has been created and linked in the description for this PR.

michael-j-green:
Also; while we're on the subject of documentation: dosbox does not appear under supported systems or cores in the documentation, so we'll need to implement that at some point.

allancoding:
Oh yeah docs...

ethanaobrien:
> Oh yeah docs...

Thanks for volunteering

PR #1095, opened 2025-09-14
Support a fixed save flushing interval

cryptonaus:
Adds the ability for the embedder to set a fixed save interval, so that users cannot. By forcing a core to flush its saves more frequently, EmulatorJS consumers can perform operations with save more reliably.

Combined with #1094, can result in frequent save update notifications.

### DIFF
--- a/data/loader.js
+++ b/data/loader.js
@@ -126,6 +126,7 @@
     config.hideSettings = window.EJS_hideSettings;
     config.browserMode = window.EJS_browserMode;
     config.shaders = Object.assign({}, window.EJS_SHADERS, window.EJS_shaders ? window.EJS_shaders : {});
+    config.fixedSaveInterval = window.EJS_fixedSaveInterval;
 
     let systemLang;
     try {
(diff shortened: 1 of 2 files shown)


michael-j-green:
Is there such a thing as too often?

For example; if I set it to flush every second (because someone will), will the cores have an issue with this? Will there be performance issues?

cryptonaus:
I initially set mgba to 80 milliseconds, which was fine, though I did just try setting it to 1 millisecond. It brought my CPU usage up to 100%, but the game is still running at full speed. (I can see the two threads separately). There almost certainly could be too much saving—and it likely is up to the core—but I'm not sure how often you would have to do that. It is probably a function of how large the save file is, more than anything else.

The other alternative I thought about for implementing this was changing the timer so that it was a delay between flushes—but I don't know if the C `cmd_savefiles` is actually synchronous or not.

ethanaobrien (approved):

michael-j-green:
Looks good. Please provide proposed documentation, and I'll approve :)

cryptonaus:
Proposed documentation 👍

michael-j-green:
Thankyou @cryptonaus

cryptonaus:
Glad I could contribute. Thank you two!

PR #1105, opened 2025-10-11
Improve ROM loading UX

298A-E9E3:
Make cores that require threads only available if threads are enabled
Make ROMs whose extensions are the name of a core use that core (i.e. rom.psx would use the psx core, and rom.atari2600 would use the atari2600 core)

### DIFF
--- a/index.html
+++ b/index.html
@@ -205,9 +205,19 @@
                             "Commodore VIC20": "vice_xvic",
                             "Commodore Plus/4": "vice_xplus4",
                             "Commodore PET": "vice_xpet",
-                            "DOSBOX-PURE": "dosbox_pure"
+                        }
+                        
+                        if (enableThreads) {
+                            coreValues["DOSBOX-PURE"] = "dosbox_pure";
+                            coreValues["PlayStation Portable"] = "ppsspp";
                         }
 
+                        for (let core in coreValues) {
+                            if (core.toLowerCase() === ext) {
+                                resolve(core)
+                            }
+                        }
+ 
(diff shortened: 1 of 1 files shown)


michael-j-green:
~~I like the idea of checking if a core supports threads before executing. That way we can fail early.~~

~~However, the execution of this should use the `cores.json` file (see https://cdn.emulatorjs.org/latest/data/cores/cores.json). There is a boolean option called `requireThreads`.~~

~~Edit: a copy of this `cores.json` file specific to the core is included with the WASM files.~~

~~That would allow EJS to detect the requirement without having to make a code change.~~

~~I believe we make the same detection for `supportsMouse` and `useKeyboard`.~~

~~Edit: I know line 46 of `emulator.js` does have an array of cores that requires threads - I think that might be a left over from someones testing. I believe the correct place to be looking is in `defaultCoreOpts` (see line 608) of `emulator.js`.~~

~~If you like, I can take a look at implementing this later tonight my time (it's morning here in Sydney).~~

michael-j-green:
~~Regarding the extension based core selection; please move that to another PR. Try to keep PR's focused on one bug fix or feature at a time.~~

~~This keeps the changes you want to implement far simpler, easier to read, and easier to test, which in turn helps prevent unintentional side-effects and bugs.~~

michael-j-green:
I've striked out my comments, because I didn't read the change you were making properly. I didn't realise you were only updating the `index.html` file. Apologies.

ethanaobrien (approved):

PR #1106, opened 2025-10-12
Refactor checkCompression to cache decompression results

michael-j-green:
Presented as a proof of concept, is caching decompressed objects into an indexedDB database for faster loading of the emulator on subsequent loads.

By modifying the `checkCompression` method, all compressed files are included in cache management by default - this includes cores, rom files, bios files, etc.

Please don't judge the code quality or location too harshly - it was vibe coded using GitHub Copilot. A proper production implementation would require a code cleanup, refactoring, and options to configure the cache (bypass, max size, eviction criteria, etc). This initial version is very much a proof of concept.


ethanaobrien:
Is this a draft or ready for review? What am I reviewing?

michael-j-green:
This is very much a draft, meant for reviewing the idea, and possibly the implementation (in terms of using an IndexedDB to store uncompressed data).

This is a proof of concept to create discussion mostly.

I do think there are benefits to this. While the initial load takes as long as it currently does, future loads are almost instant. Which is a benefit to users of the MAME or FBNEO cores as they're quite large.

I couldn't think of an easy way to share the code with you so you could try it out while also allowing discussion. Happy to kill the PR and move somewhere else if that helps.

ethanaobrien:
I think here is fine. I do wonder however about the goal of the implementation

- Why is there a 30 day timer?
- I don't think this feature should rely on the `window.crypto` api seeing how usage was [rejected in the past](https://github.com/EmulatorJS/EmulatorJS/pull/1094) due to compatibility concerns with http (many users use this an an internal network project and don't bother securing it with https.)
- I don't see a need to add another frontend for the indexeddb api when we already have one via [storage.js](https://github.com/EmulatorJS/EmulatorJS/blob/main/data/src/storage.js) (also see [here](https://github.com/EmulatorJS/EmulatorJS/blob/54d7dfbc1ff3bedd415e7fe5050928c65197717a/data/src/emulator.js#L330-L342)). If more methods need to be added we can add more methods

michael-j-green:
I totally admit this needs to be refactored :) It was only ever meant to prove that it could work.

My thinking was that if you liked the idea, I would revert the code in this PR back to the current code base and re-write it to be more compliant with the rest of the EJS code base. So, the code as presented is completely disposable. It's just there to prove the idea.

To respond to your callouts:
- the 30-day timer (time period would be configurable) is there to purge unused cache objects. The reason for this is that according to documentation around indexedDB, the total available storage space can be from a few MB to 60% of the total disk size (see https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria for more details).  I don't want a full cache to trigger a random eviction of cache objects - this could have a negative effect on user experience.
- I was aware of that change (couldn't remember which change invoked it though). The production version would attempt to leverage the solution in that PR as a common component.
- Agreed on the storage front. This was a quick and dirty "make it work" moment.

Hope this helps :)

michael-j-green:
Thought I'd attach a video to show performance. First run is with an empty cache, second is with the two objects (core and ROM file) cached.

https://github.com/user-attachments/assets/2e3a2e31-b379-4d5e-add4-aa7de9287555

ethanaobrien:
We'll also want to take into note, when we cache extracted files we shouldnt also cache the ROM/bios/etc separately, as we have now (since this stores x2 the rom size, if not more than)

michael-j-green:
It's not a terrible thing to cache the ROMs as for example if you're using a PS1 ISO - that's a lengthy decompress. Especially on mobile hardware.

Perhaps in the caching options, make it selectable what content types can be cached?

```javascript
EJS_cachingOptions = {
  "expiryDays": 30,
  "cacheROMFiles": true,
  "cacheCores": true,
  "cacheBios": true,
  "cacheOthers": true
}
```
`cacheOthers` would cover files presented by `EJS_externalFiles`

The issue with the current caching is that it's caching the compressed object, which already seems redundant since that should be cached by the browser right?

ethanaobrien:
> The issue with the current caching is that it's caching the compressed object, which already seems redundant since that should be cached by the browser right?

This is what I was trying to say

michael-j-green:
Ah fair enough. Sometimes it's hard to convey through text :)

I'll have a crack at refactoring all of this to minimise redundancy, make it effective, etc.

Give me two weeks as I'll be unavailable next week (starting Saturday) to get something built.

kuma22212:
Sorry, but I'd like to be able to download the game.
Please consider

kuma22212:
Zip compression is also fine

michael-j-green:
@kuma22212; sorry, what are you asking for? I’m confused about why you’re asking to download a game on a PR that is about caching the decompressed content for faster start times?

michael-j-green:
@ethanaobrien; here's my latest version of this. See https://github.com/EmulatorJS/EmulatorJS/blob/98ff7a172516d89fa465bf5548050f3ce0c85e63/CACHING.md for a run down on how it all works.

# Quick high level of the changes:
- removed all of the existing caching code
- added new classes to manage the cache - which in turn uses the existing storage class
  - cache is configurable to limit the max size and eviction criteria (clean up rule is essentially Least Recently Used (LRU)):
    - defaults to a max of 4GB
    - max item age of 5 days
  - only the decompressed versions of files are stored in the cache - rely on the browser to handle actual download caching
- cores, roms, bios files are all cached
- when the emulator starts:
  - download the source file (this is to check if it's changed) using in order the following - this respects the browsers caching configuration:
    - HEAD call to the server to get last changed
    - fall back to GET if HEAD is not supported
  - perform a basic hash on the downloaded source file and compare to what's in the indexedDB
    - if the hash matches an existing object in the cache, the lastAccessed attribute of the cache object is updated, and the byte arrays of all the files is returned
    - if the hash doesn't match an existing object, or there is no existing object, the downloaded source file is added to the cache and the byte arrays of all the files is returned
  - 5 seconds after emulator start up a cleanup method is called which checks the ages of all objects in the cache and expires them out of the cache - manual cache cleanup is available in the Cache Manager
  - loading times are added to the console - as these changes progress to production ready code I'll probably put these behind the debug flag

# Some basic benchmarks:
Starting a non-cached instance of the NES emulator to play Super Mario Bros:
image

Starting the same ROM and core from cache:
image

# IndexedDB database
image

# User interface change
image

### Notes

- file type is determined by the method used to upload it (for example: `EJS_gameUrl` will set the type to `ROM`)
- "Clear All" will delete all as usual
- "Cleanup Now" will trigger the cache cleanup method

# Still to do

- localisation for new labels in the Cache Manager
- more vigorous testing with large caches and ROMs (PlayStation is probably a good candidate for this)
- files added by `EJS_externalFiles` are not currently included - I'm not sure I want to include this type as the content could be dynamic and not suitable for caching.

ethanaobrien:
This method works, is this ready for review?

If you'd like to cleanup the methods of downloading/decompressing some more, at some point I would like to replace all instances of downloadFile with a class that supports various `toText`, `toJson`, `toBlob`, `toUint8Array`, etc (whether or not to check for decompression provided and set in the constructor.

michael-j-green:
I think it's ready for review.

I do like that idea. The current downloadFile code is quite complex, and not easy to follow, so breaking it up into smaller chunks would be ideal - though I think I'd like to implement that as a separate PR as this one is already quite large, and I'm loathe to make large sweeping changes in one go.

michael-j-green:
Hi @ethanaobrien, are you able to review this? I've also added @S-Coding23 as a secondary reviewer.

ethanaobrien:
I can later this week

S-Coding23:
> Hi @ethanaobrien, are you able to review this? I've also added @S-Coding23 as a secondary reviewer.

Thank you, I'll have some time tomorrow to review.

michael-j-green:
@ethanaobrien; @S-Coding23; any joy with reviewing this one? :)

ethanaobrien:
fiiiiine

data/src/cache.js
```diff
@@ -0,0 +1,232 @@
...
+     * @param {number} maxAgeMins - Maximum age of items (in minutes) before they are cleaned up.
+     */
+    constructor(enabled = true, storage, maxSizeMB = 4096, maxAgeMins = 7200) {
+        this.enabled = enabled;
+        this.storage = storage;
+        this.maxSizeMB = maxSizeMB;
+        this.maxAgeMins = maxAgeMins;
+        this.minAgeMins = Math.max(60, maxAgeMins * 0.1); // Minimum 1 hour, or 10% of max age
+
+        console.log('Initialized EJS_Cache with settings:', {
```
ethanaobrien: Usually all log statements are only printed if debug is true. What do you think about this?

data/src/cache.js
```diff
@@ -0,0 +1,232 @@
...
+
+        // check if the item exists, if so remove the existing item
+        const existingItem = await this.get(item.key);
+        if (existingItem) {
+            await this.storage.remove(item.key);
+        }
+
+        // check that the size of item.files does not cause the cache to exceed maxSizeMB
+        let currentSize = 0;
+        const allItems = await this.storage.getAll();
```
ethanaobrien, line 443: Does this load the entire database into RAM or do you know if the browser only reads it on an attempt to access it?

data/src/emulator.js
```diff
@@ -329,16 +329,23 @@ class EmulatorJS {
...
-                core: new window.EJS_DUMMYSTORAGE()
+                // Remove rom and bios storage - rely on browser cache for files and checkCompression for decompression
             }
+            this.storageCache = new window.EJS_Cache(false, new window.EJS_DUMMYSTORAGE(), 1, 1);
         } else {
             this.storage = {
-                rom: new window.EJS_STORAGE("EmulatorJS-roms", "rom"),
-                bios: new window.EJS_STORAGE("EmulatorJS-bios", "bios"),
-                core: new window.EJS_STORAGE("EmulatorJS-core", "core")
+                // Remove rom and bios storage - rely on browser cache for files and checkCompression for decompression
```
ethanaobrien: I'm not sure how I feel about this

https://support.ptc.com/help/thingworx/platform/r10.0/en/index.html#page/ThingWorx/Help/Composer/System/Subsystems/BrowserConfigurationsForCachingInWindows.html

> Google Chrome can cache files up to 130 MB.
> Mozilla Firefox can cache files sizes up to 180 MB.

Caching also isn't persistent per-device or browser, I fear it could lead to unnecessary load times or an extra amount of bandwidth used.

Also in my tests, the core `.data` files are never getting cached by the browser.

This is also highly dependent on the website owner to properly set their cache-control preferences, which can be a pain if you only want it for a few files and not others.

data/src/emulator.js
```diff
@@ -329,16 +329,23 @@ class EmulatorJS {
...
+            this.storageCache = new window.EJS_Cache(true, new window.EJS_STORAGE("EmulatorJS-cache", "cache"), this.config.cacheMaxSizeMB || 4096, this.config.cacheMaxAgeMins || 7200);
+            
+            // Run initial cleanup after cache initialization (non-blocking)
+            setTimeout(async () => {
+                try {
+                    await this.storageCache.cleanup();
+                } catch (error) {
+                    console.error('[EJS Cache] Error during startup cleanup:', error);
+                }
+            }, 5000); // 5 second delay to avoid blocking startup
```
ethanaobrien: Instead of having a delay, why not just run this at the end of the contructor?

data/src/emulator.js
```diff
@@ -559,16 +566,128 @@ class EmulatorJS {
...
+                const cacheStoreStartTime = performance.now();
+                const fileItems = [];
+                
+                // Use collected files if callback was used, otherwise use returned files
+                const filesToCache = callbackWrapper ? collectedFiles : decompressedFiles;
+                
+                for (const [filename, fileData] of Object.entries(filesToCache)) {
+                    if (fileData && fileData !== true) {
+                        fileItems.push(new window.EJS_FileItem(filename, fileData));
+                        console.log(`[EJS Cache] Adding file to cache: ${filename} (${fileData ? fileData.byteLength || fileData.length || 'unknown size' : 'no data'} bytes)`);
```
ethanaobrien: Please use double quotes

data/src/emulator.js
```diff
@@ -604,21 +723,21 @@ class EmulatorJS {
             console.warn("Threads is set to true, but the SharedArrayBuffer function is not exposed. Threads requires 2 headers to be set when sending you html page. See https://stackoverflow.com/a/68630724");
             return;
         }
-        const gotCore = (data) => {
+        const gotCore = (data, shouldCacheDecompressed = false, baseCoreId = null) => {
```
ethanaobrien: What are these new flags used for?

data/src/emulator.js
```diff
@@ -636,11 +755,41 @@ class EmulatorJS {
...
+                // The core decompression is now handled by checkCompression which already caches the result
+                // No need for additional core-specific caching - this would create duplicates
+                console.log(`[EJS Core] Core decompression complete (cached by checkCompression)`);
+
                 this.initGameCore(js, wasm, thread);
             });
         }
+
+        // Helper function to generate cache key for core data
+        const generateCoreDataCacheKey = (data) => {
```
ethanaobrien: This is the third time this one function is in this file, I think we should make a utils file of some sort.

data/src/emulator.js
```diff
@@ -636,11 +755,41 @@ class EmulatorJS {
...
+            const now = Date.now();
+            const ageMins = (now - cachedItem.lastAccessed) / (1000 * 60);
+            // Use the same expiration logic as the cache (7200 minutes = 5 days)
+            const maxAgeMins = this.storageCache.minAgeMins || 7200;
+            return ageMins > maxAgeMins;
+        };
         const report = "cores/reports/" + this.getCore() + ".json";
-        this.downloadFile(report, null, false, { responseType: "text", method: "GET" }).then(async rep => {
+        // Add cache-busting parameter periodically to ensure we get updated build versions
+        // This ensures that when cores are updated, we'll eventually get the new buildStart value
```
ethanaobrien, line 753: I thought we added the build reports with the build date to pull according to the json...

Server bandwidth even with our current caching method averages 50% of the max bandwidth. I'm not sure how I feel about updating it every hour (in most cases, why would a user be loading multiple pages of the same core in an hour that it makes that much of a difference)

data/src/emulator.js
```diff
@@ -668,13 +817,56 @@ class EmulatorJS {
...
+                    }).catch(() => null);
+                    
+                    if (quickDownload && quickDownload.data) {
+                        // Generate cache key the same way checkCompression does
+                        const dataArray = new Uint8Array(quickDownload.data);
+                        let hash = 0;
+                        for (let i = 0; i < dataArray.length; i++) {
+                            hash = ((hash << 5) - hash + dataArray[i]) & 0xffffffff;
+                        }
+                        const compressionCacheKey = `compression_${hash}_${dataArray.length}`;
```
ethanaobrien: fourth time 😭

docs/CACHING.md
```diff
@@ -0,0 +1,250 @@
+# EmulatorJS Caching System
```
ethanaobrien, line 1: I think this should be moved to a docs or development folder

test_rom_cache.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
```
ethanaobrien: Should go in tests folder

ethanaobrien (changes requested):
Might be a better idea (and clean up a lot) if we combined the fetch functions with caching. Call fetch everywhere we need it, and automatically call the function for caching and/or decompression.

S-Coding23:
Overall this looks awesome, but I think it would be very helpful if the caching system was exposed so it can be tuned per environment. Something like this: 

EJS_Cache.enabled = true;
EJS_Cache.maxSizeMB = 500; 
EJS_Cache.maxAgeDays = 14;

data/src/cache.js
michael-j-green: Agreed... I hadn't done this for the first pass, but will wrap them all.

michael-j-green: I think I've got them all now...

data/src/cache.js
michael-j-green: Good callout, it did load all everything into memory.

I have now moved cached blobs to their own database. They have the same keys as the metadata, and are loaded and deleted at the same time. This should optimise memory usage, only loading the blobs when needed.

I've also updated the cache cleanup method to remove blobs that have a key not present in the cache table.

test_rom_cache.html
michael-j-green: Just going to delete it altogether. This was an oversight.

data/src/emulator.js
michael-j-green: That was done to keep cache configuration together in the code to hopefully make it a bit easier to follow. Didn't want to be jumping all over the place :)

Also; you can't async on the constructor, so it would be a blocking action. Happy to move it, but since the goal of this PR is to improve start time, I think leaving it as a background task that executes while the user is going through the start up phase of their game is best (that is; it'll run while the credits are being displayed in GoldenEye for example).

data/src/emulator.js
michael-j-green: Fixed this and all other mal-formatting. I had been using a new PC, and forgot to disable VSCode's formatter.

data/src/emulator.js
michael-j-green: Were used during development. They're now gone.

data/src/emulator.js
michael-j-green: I think so too... throw the hasher in there. I'll have a look around for some good candidates for moving into a utils class once I've got the majority of caching sorted :)

data/src/emulator.js
michael-j-green: Agreed... I had actually forgotten about the build reports file :D

I'm thinking of moving all of the download functions into their own class to hopefully make things a bit easier to read. What do you think?

data/src/emulator.js
michael-j-green: I've moved all the hashing algorithms to utils.js and anytime I encountered a hashing function, I pointed it there.

data/src/emulator.js
ethanaobrien: Ah, sounds good then. There is no such thing as a non-async blocking action in javascript

I do think this should be reduced though, it should be fine even if it was 0

data/src/emulator.js
ethanaobrien: >I'm thinking of moving all of the download functions into their own class to hopefully make things a bit easier to read. What do you think?

This would be amazing! Having this download function handle caching too would also prevent many duplicated database checks

data/src/emulator.js
ethanaobrien: Looks good, thanks

data/src/emulator.js
michael-j-green: I think 0 might be a bit too low. There's the possibility of hitting cleaning the cache while EJS is trying to read it.

When the cache is read, it reads the metadata and updates the lastAccessed property, before loading the blob from it's table. That's three steps, so there is the (very small) potential for the cleaner to prune the cache before the lastAccessed property is updated, which would make the read operation fail due to the blob being missing.

Now granted if the cache returns null or undefined, the checkCompression method should see this as a cache fail and redownload the content.

Might set it to 100. The majority of processing should be done by then.

data/src/emulator.js
michael-j-green: Sweet... I'll move them to the new utils class.

Additional upside: `emulator.js` becomes more about running the emulator than handling files :)

michael-j-green:
> Overall this looks awesome, but I think it would be very helpful if the caching system was exposed so it can be tuned per environment. Something like this:
> 
> EJS_Cache.enabled = true; EJS_Cache.maxSizeMB = 500; EJS_Cache.maxAgeDays = 14;

Thanks! Was planning on doing this once I got everything bedded down :)

data/src/emulator.js
ethanaobrien: 100 would work

>There's the possibility of hitting cleaning the cache while EJS is trying to read it.

The issue here is for cases like startOnLoad or for slower devices. Any number could lead to a race condition in this case, it might just be better to handle the off chance it gets deleted the moment we try to pull it

michael-j-green: Had what might be a better idea...

1. create a `startupCleanupCompleted` flag - default to false
2. at the start of the `get` method, check if `startupCleanupCompleted` === false
3. if false, run the cleanup (it needs to be run async, and `get` is called async)
4. set `startupCleanupCompleted` to true.

That way it's always called async, and it's run at a time that is convenient.

michael-j-green:
Added the `EJS_cacheConfig` option to configure the cache manager.

This option takes an object and is configured thusly:
```javascript
EJS_cacheConfig = {
    enabled: true,
    cacheMaxSizeMB: 4096,
    cacheMaxAgeMins: 7200
};
```

Note that the option `EJS_CacheLimit` no longer does anything and has been removed (need to make a note in the documentation).

I did find another option called `EJS_disableDatabases` which is not documented, which seems designed to disable creating databases on the browser. With the addition of the caching, this option now does nothing and has been removed.

data/src/emulator.js
michael-j-green: > Had what might be a better idea...
> 
> 1. create a `startupCleanupCompleted` flag - default to false
> 2. at the start of the `get` method, check if `startupCleanupCompleted` === false
> 3. if false, run the cleanup (it needs to be run async, and `get` is called async)
> 4. set `startupCleanupCompleted` to true.
> 
> That way it's always called async, and it's run at a time that is convenient.

Ended up migrating to this method. It works well!

data/src/storage.js
```diff
@@ -100,11 +81,33 @@ class EJS_STORAGE {
...
+            if (!keys) return resolve([]);
+            resolve(keys);
+        });
+    }
 }
 
 class EJS_DUMMYSTORAGE {
     constructor() {}
-    addFileToDB() {
+    #addFileToDB() {
```
ethanaobrien: While I like the idea of using private functions - this is relatively new https://caniuse.com/mdn-javascript_classes_private_class_methods

This wouldn't be an issue, though I have concerns since I've had issues for [replaceAll](https://caniuse.com/wf-string-replaceall) not being supported on their browser either

ethanaobrien (changes requested):

data/src/storage.js
michael-j-green: Hey; just getting back into this. Had writers block trying to untangle the download routines.

Happy to make the private functions public to remove that limitation. Before I do though; is there a minimum browser version we want to be targetting? Might be worthwhile deciding on that and publishing it in the README so that future contriutions can keep that in mind when submitting changes :)

I'm thinking broadly we should target:
- Chromium and Safari as first class citizens
- Firefox as best effort (given it's WASM performance issues)
- Any other browser will be best effort, and any requests will be judged on a case by case basis, though serious weight will be added if that request benefits the above two
- Internet Explorer will not be considered at all

We should specify the minimum version for the browser that we'll support, and in the case of mobile OS's (where the browser version is often tied to the OS) we should specify the minimum Android and iOS version we support.

Browser version market share can be found here: https://gs.statcounter.com/browser-version-market-share as well as a bunch of other metrics to help us decide the minimum version we want to support.

michael-j-green:
@ethanaobrien: alright; I think I've disentangled most of the download calls, and centralised all decompression routines into EJS_Download.

Would you be able to run a quick eye over it? I know it needs a bit more clean up, but a) it's working now, and b) I need sleep :D

~~Once you're happy, I'll resolve the merge conflicts that I know are going to be incoming.~~ Nevermind: it was a smaller fix than I expected :D

Edit: nevermind again... turns out there were more download methods to disentangle.

michael-j-green:
Looks like I've made some pretty good progress on the caching and download command cleanup. Still a fair bit of work to do, but it's coming along nicely.

To start with, I've tried to classify the type of download:

```javascript
downloadType = {
    "rom": { "name": "ROM", "dontCache": false },
    "core": { "name": "Core", "dontCache": false },
    "bios": { "name": "BIOS", "dontCache": false },
    "parent": { "name": "Parent", "dontCache": false },
    "patch": { "name": "Patch", "dontCache": false },
    "reports": { "name": "Reports", "dontCache": false },
    "states": { "name": "States", "dontCache": true },
    "support": { "name": "Support", "dontCache": true },
    "unknown": { "name": "Unknown", "dontCache": true }
}
```

This lives at the top of the emulator.js file to make it easy to edit if need be.

The `name` is more of a display name for the purposes of logging, and `dontCache` is a flag to prevent the caching.

For example; we probably don't want save states to be cached.

Both url and HTML File object provided files are catered for in the new `download` method.

This method determines the input type (string url where we have to reach out to a server to retrieve a file, or File object provided by `index.html`).

For a File object, it generates a key based on the provided data and checks the cache for that key. If present it sets an internal return value (more on that later). If it's not present in the cache it extracts it and adds it to the cache, then puts the resultant EJS\_CacheItem into the internal return value.

For a url object, it calls another new method called `downloadFile`. This does some data massaging before calling the `download` method inside `EJS_Download`. The `EJS_Download` method does the cache checking and returning as well as extraction, etc. Once a return value is created (either through download or cache return), the `download` method sets the internal variable to the returned EJS\_CacheItem object.

Now we have that object, `download` will write all of the files within the object (which are decompressed at this point) to the emscripten file system.

So the one method `download` handles downloading (including cache handling) and writing to the file system.

I've updated the `downloadFiles` method with the following:

```javascript
const romData = await this.download(this.config.gameUrl, this.downloadType.rom);
const biosData = await this.download(this.config.biosUrl, this.downloadType.bios);
await this.downloadStartState();
const parentData = await this.download(this.config.gameParentUrl, this.downloadType.parent);
const patchData = await this.download(this.config.gamePatchUrl, this.downloadType.patch);
```

These replace `downloadRom`, `downloadBios`, `downloadGameParent`, and `downloadGamePatch`.

This method has also had the `selectRomFile` method added to it, as the final step to setting up the emulator before calling `this.startGame();`.

It still needs more testing and a load of clean up work. But it's made a lot of the code reusable.

I also made some updates to the `index.html` page now that all downloads can be provided as File objects

image

Clicking the "Advanced Options" checkbox results in the following:

image

From here you can provide a BIOS file, enable debug, enable threads (the checkbox doesn't appear if SharedArrayBuffer isn't available), and set the browser mode (auto, mobile, or desktop - auto is the default).

The debug, threads and browser mode options are pre-set if your url query string enables them.

data/src/storage.js
michael-j-green: As per our conversation, the new minimum supported version of Javascript will be ES2020.

I've added support and configuration files that install ESLinter to flag any non-compliant code, as well as an action to prevent merging of non-compliant code.

Users of VSCode will also now be suggested to install the ESLint extension so that non-compliant code is called out as an error:
Screenshot 2025-12-18 at 9 59 42 am

Should the need arise, the Ecmascript version that ESLint is looking for can be changed in the `eslint.config.js` file.

michael-j-green:
As per our conversation, the new minimum supported version of Javascript will be ES2020.

I've added support and configuration files that install ESLinter to flag any non-compliant code, as well as an action to prevent merging of non-compliant code.

Users of VSCode will also now be suggested to install the ESLint extension so that non-compliant code is called out as an error:
Screenshot 2025-12-18 at 9 59 42 am

Should the need arise, the Ecmascript version that ESLint is looking for can be changed in the `eslint.config.js` file.

michael-j-green:
@ethanaobrien; I've taken another pass at this PR, and I'm reasonably happy it meets MVP status.

Would you be able to run an eye over it and make sure you're happy with it?

If you are, and I think given the sheer amount of code changes this PR includes, I think I might ask some of the other users of EJS to test this code to make sure I've covered all of the use cases.

Goals are: it should work exactly as it did before with no code changes for users of the project - only faster.

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+     * @param {boolean} forceExtract - Whether to force extraction of compressed files regardless of extension (default is false).
+     * @param {boolean} dontCache - If true, the downloaded file will not be cached (default is false).
+     * @returns {Promise<EJS_CacheItem>} - The downloaded file as an EJS_CacheItem.
+     */
+    downloadFile(url, type, method = "GET", headers = {}, body = null, onProgress = null, onComplete = null, timeout = 30000, responseType = "arraybuffer", forceExtract = false, dontCache = false) {
+        let cacheActiveText = " (cache usage requested)"
+        if (dontCache) {
+            cacheActiveText = "";
+        }
+        console.log("[EJS Download] Downloading " + responseType + " file: " + url + cacheActiveText);
```
ethanaobrien, line 107: I think logs should go through a function in the class that'll append that `[EJS Download]` or similar tag to the beginning of the argument list. It should also not print if debug is off

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                    const cacheControl = resp.headers.get("Cache-Control");
+                    const expires = resp.headers.get("Expires");
+                    if (cacheControl && /max-age=(\d+)/.test(cacheControl)) {
+                        const maxAge = parseInt(cacheControl.match(/max-age=(\d+)/)[1]);
+                        cacheExpiry = now + maxAge * 1000;
+                    } else if (expires) {
+                        const exp = Date.parse(expires);
+                        if (!isNaN(exp)) cacheExpiry = exp;
+                    } else {
+                        cacheExpiry = now + 5 * 24 * 60 * 60 * 1000;
```
ethanaobrien, line 185: Please add comments for what time range this is

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                    }
+                    if (responseType === "arraybuffer") {
+                        const contentLength = parseInt(resp.headers.get("Content-Length") || "0");
```
ethanaobrien: This should probably be changed to have the `|| 0` outside of the parseInt block. In the event that the server sends an invalid `Content-Length` header but it is still there, this would result in contentLength being `NaN`

`NaN || 0` results in `0`

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                            if (onProgress && contentLength) {
+                                onProgress("downloading", Math.floor(received / contentLength * 100), received, contentLength);
+                            }
+                        }
+                        data = new Uint8Array(chunks.reduce((acc, val) => acc + val.length, 0));
+                        let offset = 0;
+                        for (const chunk of chunks) {
+                            data.set(chunk, offset);
+                            offset += chunk.length;
+                        }
```
ethanaobrien: It might be faster/better to call the Blob constructor with the chunks, and read the arraybuffer from that. Right now we're iterating through the chunks twice, once to get the total size and once to set the data

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                        try {
+                            const compression = new window.EJS_COMPRESSION(this.EJS);
+                            await compression.decompress(data, (msg, isProgress) => {
+                                if (onProgress && isProgress) {
+                                    const percent = parseInt(msg);
+                                    onProgress("decompressing", isNaN(percent) ? 0 : percent, 0, 0);
+                                }
+                            }, (fname, fileData) => {
+                                files.push(new EJS_FileItem(fname, fileData instanceof Uint8Array ? fileData : new Uint8Array(fileData)));
+                            });
```
ethanaobrien, line 238: Wouldnt it make more sense to have `EJS_COMPRESSION.decompress` callback with an `EJS_FileItem`?

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                        } catch (e) {
+                            reject(`Decompression failed: ${e}`);
+                            return;
+                        }
+                    } else {
+                        files = [new EJS_FileItem(filename, data instanceof Uint8Array ? data : new Uint8Array(data))];
```
ethanaobrien, line 244: Same query here

data/src/cache.js
```diff
@@ -0,0 +1,551 @@
...
+                } else if (data instanceof Uint8Array) {
+                    files = [new EJS_FileItem(filename, data)];
+                } else if (data instanceof ArrayBuffer) {
+                    files = [new EJS_FileItem(filename, new Uint8Array(data))];
+                } else {
+                    // Fallback: try to convert to string then encode
+                    const encoder = new TextEncoder();
+                    files = [new EJS_FileItem(filename, encoder.encode(String(data)))];
+                }
+                }
```
ethanaobrien: Indentation is wrong somewhere

data/src/emulator.js
```diff
@@ -89,75 +89,137 @@ class EmulatorJS {
...
-                });
-            }
-            xhr.onload = function() {
-                if (xhr.readyState === xhr.DONE) {
-                    let data = xhr.response;
-                    if (xhr.status.toString().startsWith("4") || xhr.status.toString().startsWith("5")) {
-                        cb(-1);
-                        return;
+
+            // Use EJS_Download for http(s) downloads
```
ethanaobrien: Would it not be possible to merge these?

data/src/emulator.js
```diff
@@ -327,19 +390,54 @@ class EmulatorJS {
...
-                rom: new window.EJS_STORAGE("EmulatorJS-roms", "rom"),
-                bios: new window.EJS_STORAGE("EmulatorJS-bios", "bios"),
-                core: new window.EJS_STORAGE("EmulatorJS-core", "core")
-            }
+
+        this.storage = {}
+    
+        // set cache configuration defaults
+        const cacheConfigDefaults = {
+            enabled: true,
```
ethanaobrien: This should be false if `this.config.disableDatabases` is false. No reason to remove the option when we can just map it, in my opinion, unless you think otherwise

data/src/emulator.js
```diff
@@ -606,48 +693,86 @@ class EmulatorJS {
...
+            if (this.debug) console.log("[EJS Core] js size:", js?.byteLength, "wasm size:", wasm?.byteLength, "thread size:", thread?.byteLength);
+
+            this.initGameCore(js, wasm, thread);
         }
+
         const report = "cores/reports/" + this.getCore() + ".json";
-        this.downloadFile(report, null, false, { responseType: "text", method: "GET" }).then(async rep => {
+        // Add cache-busting parameter periodically to ensure we get updated build versions
+        // This ensures that when cores are updated, we'll eventually get the new buildStart value
+        const cacheBustInterval = 1000 * 60 * 60; // 1 hour
```
ethanaobrien, line 754: Is this telling the downloader to only have the cache last for an hour?

data/src/emulator.js
```diff
@@ -993,11 +1039,97 @@ class EmulatorJS {
...
+            if (this.debug) console.log("Disable CUE handling:", disableCue);
+
+            const fileNames = [];
+            for (const file of romData.files) {
+                if (file.filename.endsWith("/")) {
+                    continue;
+                }
+                fileNames.push(file.filename);
+            }
+            selectRomFile(fileNames, this.getCore(), disableCue);
```
ethanaobrien: I'd like this block divided into another function - downloadFiles should be a barebone " setup this, this, and this" function without any real logic in it, in my experience it helps when trying to figure out the flow of how things are setup

data/src/emulator.js
```diff
@@ -1098,6 +1230,31 @@ class EmulatorJS {
...
+                                console.log(`${indent}[DIR] ${fullPath}`);
+                                listDir(fullPath, indent + "  ");
+                            } else {
+                                console.log(`${indent}${fullPath}`);
+                            }
+                        }
+                    } catch (e) {
+                        console.warn("Error reading directory:", path, e);
+                    }
+                }
```
ethanaobrien: I think we can make this a utility function

data/src/emulator.js
```diff
@@ -1278,132 +1435,133 @@ class EmulatorJS {
...
+                displayName: "Disks"
+            },
+            contextMenu: {
+                visible: true,
+                icon: '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><!--!Font Awesome Free 6.5.1 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free Copyright 2023 Fonticons, Inc.--><path d="M0 96C0 78.3 14.3 64 32 64H416c17.7 0 32 14.3 32 32s-14.3 32-32 32H32C14.3 128 0 113.7 0 96zM0 256c0-17.7 14.3-32 32-32H416c17.7 0 32 14.3 32 32s-14.3 32-32 32H32c-17.7 0-32-14.3-32-32zM448 416c0 17.7-14.3 32-32 32H32c-17.7 0-32-14.3-32-32s14.3-32 32-32H416c17.7 0 32 14.3 32 32z"/></svg>',
+                displayName: "Context Menu"
+            }
+        };
```
ethanaobrien, line 1565: I can't recall where this was mentioned, or maybe it was just a dream that you had pointed it out. If you want to go ahead and put this into a separate file similar to how shaders are done, that'd be great

data/src/emulator.js
```diff
@@ -2423,57 +2581,153 @@ class EmulatorJS {
...
                 "Close": () => {
                     this.closePopup();
                 }
             });
-            const roms = await this.storage.rom.getSizes();
+            
             list.style.width = "100%";
             list.style["padding-left"] = "10px";
             list.style["text-align"] = "left";
             body.appendChild(list);
```
ethanaobrien: Not that it matters but could we move this under the calls to append the table to list

build.js
ethanaobrien: Could you fix these warnings

index.html
```diff
@@ -258,23 +389,29 @@ <h1>EmulatorJS Demo</h1>
...
+                div.appendChild(sub);
+                advanced.remove();
                 document.body.appendChild(div)
 
                 window.EJS_player = "#game";
                 window.EJS_gameName = parts.shift();
-                window.EJS_biosUrl = "";
+                window.EJS_biosUrl = biosUrl;
                 window.EJS_gameUrl = url;
+                // window.EJS_gameUrl = "smb.zip";
```
ethanaobrien: smb.zip?

index.html
```diff
@@ -258,23 +389,29 @@ <h1>EmulatorJS Demo</h1>
...
                 window.EJS_pathtodata = "data/";
                 window.EJS_startOnLoaded = true;
                 window.EJS_DEBUG_XX = enableDebug;
-                window.EJS_disableDatabases = true;
                 window.EJS_threads = enableThreads;
                 if (browserMode) {
                     window.EJS_browserMode = browserMode;
                 }
+                window.EJS_cacheConfig = {
+                    enabled: true,
```
ethanaobrien: Maybe we should divide the ability to enable or disable core cache/game cache/etc, the reason it was disabled before on the demo was to not accidentally cache saves. Setting specific cache ages for core/bios/rom also might make sense, since cores are updated every once in a while (like with nightly cores, they are updated nightly) but a rom/bios would likely never change

Does the time of cache get updated when the player loads that same game again?

ethanaobrien (changes requested):
Very nice work, thank you

index.html
michael-j-green: Ooops that was a mistake left over from testing. Will be removed in the next commit.

index.html
michael-j-green: The lastaccessed attribute of the cache item gets updated whenever it's pulled from indexeddb.

Saves are specifically excluded from cache as per the this.downloadType array in the emulator.js constructor

data/src/cache.js
michael-j-green: While I was editing this I was thinking the same thing.

Once merged, I think I'd like to create a logging class to make life a lot easier - I don't want to implement here as it's a bit of scope creep that I don't want to worry about right now :D

data/src/cache.js
michael-j-green: Add comment for default 5 day expiry

data/src/cache.js
michael-j-green: Replaced that line with:
```
let contentLength = 0;
if (resp.headers.get("Content-Length")) {
    try {
        contentLength = parseInt(resp.headers.get("Content-Length"));
    } catch (e) {
        // swallow any errors parseing content length
    }
}
```

data/src/cache.js
michael-j-green: I didn't want to change how `EJS_COMPRESSION.decompress` works without good reason. If we want to use compression elsewhere and not return an EJS_FileItem then we're not locked into that design.

data/src/cache.js
michael-j-green: See above; I'll resolve this conversation and keep it restricted to the comment on line 162 (since they're the same thing).

data/src/cache.js
michael-j-green: Best I can tell it appears correct here - I even ran the VSCode formatter over it, and it's still the same.

data/src/emulator.js
michael-j-green: Potentially... the code changes were working as intended, and I didn't want to tempt fate until the overall design was locked in.

I'll come back to this one towards the end of the PR.

data/src/emulator.js
michael-j-green: I've added the below snippet below the cache config. This way we can handle `disableDatabases` as expected. For anyone using this option, the behaviour should remain the same.

```javascript
if (this.config.disableDatabases === true) {
    this.config.cacheConfig.enabled = false;
}
```

data/src/emulator.js
michael-j-green: Yes. This is to prevent pulling the reports file on every reload. Since the reports file doesn't update less than daily, it's ok to rely on the browser cache to return a report that could be up to an hour old. Reports are not cached by EJS_Cache.

data/src/emulator.js
michael-j-green: Agreed... I'll look at separating this out.

data/src/emulator.js
michael-j-green: Good idea.

michael-j-green: I've moved the list directory component to gamemanager.js

data/src/emulator.js
michael-j-green: I mentioned it in that other PR about moving and resizing the virtual gamepad.

I'm keen to move these sorts of consts to other files/locations, but would like to do that as a separate change - I don't want to introduce too many changes in this PR due to risk of breaking something else :D

data/src/emulator.js
michael-j-green: Done :)

data/src/cache.js
michael-j-green: I need to investigate this suggestion.

data/src/emulator.js
michael-j-green: See #1155

data/src/cache.js
michael-j-green: See #1154

data/src/cache.js
ethanaobrien: What is this catching? [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) does not throw any errors, and [`Headers.get`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get) only throws a `TypeError` if our static string is wrong (which will never occur).

As shown in the `parseInt` docs, `NaN` will be returned if an error occurred. The `Headers.get` method will return null if the header is not found.

```js
let contentLength = 0;
try {
    contentLength = parseInt(null);
} catch (e) {}
console.log(contentLength); // NaN
console.assert(contentLength === 0); // This fails
```

Using similar to the original syntax would be better:

```js
let contentLength = parseInt(null) || 0;
console.log(contentLength); // 0
console.assert(contentLength === 0); // This passes
```

The reason why I suggested the change was in the case the server sends an invalid content length header, somehow:

```js
let contentLength = parseInt("a123") || 0; // oops! Server dev left an extra character in there
console.log(contentLength); // 0
console.assert(contentLength === 0); // This passes
```

compared to the original change:

```js
let contentLength = parseInt("a123" || "0"); // "a123" is a "truthy" value, so 0 is not used.
console.log(contentLength); // NaN
console.assert(contentLength === 0); // This fails
```

data/src/cache.js
ethanaobrien: I think returning an `EJS_FileItem` should be fine, the class directly allows you to access the data/filename and methods/keys can be added as needed

data/src/cache.js
ethanaobrien: These lines are not indented enough: https://github.com/EmulatorJS/EmulatorJS/blob/21b83ca54d781cabc84e7b06259b85f3c52127f0/data/src/cache.js#L165-L177

data/src/emulator.js
ethanaobrien: So its more of a browser thing, that sounds good.

ethanaobrien:
Let me know when this is ready for another review

data/src/cache.js
ethanaobrien: Oh, you fixed this

data/src/emulator.js
ethanaobrien: Be sure to add this back into loader.js

data/src/cache.js
michael-j-green: Today's update should resolve this.
```javascript
let contentLength = 0;
if (resp.headers.get("Content-Length")) {
    try {
        const parsedContentLength = parseInt(resp.headers.get("Content-Length"));
        if (!isNaN(parsedContentLength) && parsedContentLength > 0) {
            contentLength = parsedContentLength;
        }
    } catch (e) {
        // swallow any errors parseing content length
    }
}
```

data/src/cache.js
michael-j-green: This is a little out of my wheelhouse, so I asked Copilot and it came up with this to replace that code:
```javascript
const blob = new Blob(chunks);
const ab = await blob.arrayBuffer();
data = new Uint8Array(ab);
```

I asked it what the drawbacks to using this method over the old method was. Here's it's response:
> ### Drawbacks
> Peak memory: Creates a Blob and then allocates a full ArrayBuffer; with large downloads this can momentarily increase memory pressure compared to allocating one Uint8Array and copying directly.
> Extra allocation: The Blob object is transient and then arrayBuffer() allocates the final buffer; this is one more allocation than a direct preallocated Uint8Array path.
> Async conversion: blob.arrayBuffer() is async; negligible overhead, but it's an extra promise hop versus a synchronous copy loop.
> Platform limits: Some browsers (especially mobile Safari) have stricter memory limits; very large Blob operations may fail or be slower than manual copying.
> No zero-copy: Still copies data into a new buffer; if Content-Length is known, a single-pass preallocated Uint8Array can avoid the intermediate Blob.
> ### Benefits
> Less JS work: Removes your explicit two-pass iteration; native Blob concatenation is usually well-optimized.
> Simplicity: Smaller, clearer code with the same resulting Uint8Array.
> ### If you want maximum efficiency
> Known size path: When Content-Length is present, preallocate new Uint8Array(contentLength) and fill it in one pass while streaming; fall back to Blob when size is unknown.
> Guard for huge files: Keep the current Blob approach for typical sizes, but consider a heuristic to switch to direct copy for very large totals to reduce peak memory.

I'm commit the updated version, but thought it was worth calling out the above notes.

build.js
michael-j-green: These are now arrow functions which suppresses the warnings.

data/src/emulator.js
michael-j-green: I believe I've separated this out sufficiently.

data/src/emulator.js
michael-j-green: This has now been reduced to a helper function, useful for standardising the progress and download completed functions, and handles the response from the cache depending on content type.

michael-j-green:
> Let me know when this is ready for another review

Ready when you are :) I think that just about covers everything.

data/src/cache.js
```diff
@@ -133,12 +133,10 @@ class EJS_Download {
                                 onProgress("downloading", Math.floor(received / contentLength * 100), received, contentLength);
                             }
                         }
-                        data = new Uint8Array(chunks.reduce((acc, val) => acc + val.length, 0));
-                        let offset = 0;
-                        for (const chunk of chunks) {
-                            data.set(chunk, offset);
-                            offset += chunk.length;
-                        }
+                        // Optimize concatenation by leveraging Blob to avoid double iteration
```
ethanaobrien: No need for this comment

data/src/emulator.js
```diff
@@ -1013,109 +1013,149 @@ class EmulatorJS {
...
+
+        if (coresThatNeedCueHandling.includes(this.getCore()) && this.config.disableCue === undefined) {
+            disableCue = true;
+        } else {
+            disableCue = this.config.disableCue;
+        }
+
+        if (this.debug) console.log("Disable CUE handling:", disableCue);
+        return disableCue;
+    }
```
ethanaobrien, line 1048: What does this function do if we do nothing with the return value

data/src/emulator.js
```diff
@@ -1013,109 +1013,149 @@ class EmulatorJS {
...
+     * Download all game-related files (ROM, BIOS, parent, patch, and start state)
+     */
+    async downloadGameFiles() {
+        const romData = await this.download(this.config.gameUrl, this.downloadType.rom);
+        const biosData = await this.download(this.config.biosUrl, this.downloadType.bios);
+        await this.downloadStartState();
+        const parentData = await this.download(this.config.gameParentUrl, this.downloadType.parent);
+        const patchData = await this.download(this.config.gamePatchUrl, this.downloadType.patch);
+        return romData;
+    }
```
ethanaobrien: I don't think I like this implementation - its a little confusing. Why are we only returning the romData? Why is everything else saved to a variable then never referenced?

ethanaobrien (changes requested):

data/src/cache.js
michael-j-green: Good catch! Removed.

data/src/emulator.js
michael-j-green: It's a remnant from the previous implementation. this.download is writing the downloads to the virtual file system (see line: 1006). I'll remove the variable assignments.

romData is returned to downloadFiles and then used to launch the emulation (this.startGameFromDownload on line 1159).

But then, this whole function is just redundant. I'll merge this in with the `downloadFiles` function.

data/src/emulator.js
michael-j-green: Oversight. I've added it into `selectRomFile` at line 1062. It replaces the previous conditional of `if (this.config.disableCue === true) {`.

michael-j-green:
Hi @ethanaobrien; just wanted to check with you if there is anything else you can see that needs fixing?

ethanaobrien:
Sorry I havent had time to review..

ethanaobrien (dismissed):
I think it is good

ethanaobrien:
hmm is there someone else to test this

michael-j-green:
Maybe the Romm guys might like to give it a crack?

I'll raise it with them.

data/src/cache.js
```diff
@@ -0,0 +1,635 @@
...
+                try { data = JSON.parse(data) } catch(e) {}
+            }
+            
+            // Clean up blob URLs to free memory
+            if (url.startsWith("blob:")) {
+                URL.revokeObjectURL(url);
+            }
+            
+            // Create a cache item for consistency
+            const filename = url.split("/").pop() || "downloaded.bin";
```
gantoine, line 58: replacing all filenames with `downloaded.bin`?

data/src/cache.js
```diff
@@ -0,0 +1,635 @@
...
+        this.debug = debug;
+
+        this.utils = new EJS_UTILS();
+
+        /**
+        * Indicates whether the startup cleanup has been completed.
+        */
+        this.startupCleanupCompleted = false;
+
+        if (this.debug) {
```
gantoine, line 314: is this approach of `this.debug` + `console.log` typical for the codebase, instead of using `console.debug`? legit question cause i have no clue lol.

data/src/GameManager.js
```diff
@@ -87,34 +87,29 @@ class EJS_GameManager {
...
-                    await new Promise(done => {
-                        this.EJS.downloadFile(this.EJS.config.externalFiles[key], null, true, { responseType: "arraybuffer", method: "GET" }).then(async (res) => {
-                            if (res === -1) {
-                                if (this.EJS.debug) console.warn("Failed to fetch file from '" + this.EJS.config.externalFiles[key] + "'. Make sure the file exists.");
-                                return done();
-                            }
+                    await new Promise(async (done) => {
+                        try {
+                            const url = this.EJS.config.externalFiles[key];
+                            const cacheItem = await this.EJS.downloadFile(url, this.EJS.downloadType.support.name, "GET", {}, null, null, null, 30000, "arraybuffer", false, this.EJS.downloadType.support.dontCache);
```
gantoine, line 93: this gives me soooo much anxiety 🙈

data/src/cache.js
michael-j-green: This is an OR.

If it can't get the filename from the url, it's using "downloaded.bin" as a default file name. It does this because the cache index and blob storage are in different tables to ensure we're only loading what's needed into memory, so we need a name to make sure we're looking up the right content.

It should be noted that in my testing I never ended up in a situation where the name was missing.

data/src/cache.js
michael-j-green: It's because we only want to add to the console if we're in debug mode. This is the current pattern.

I have an issue logged to refactor how logs are done (https://github.com/EmulatorJS/EmulatorJS/issues/1154). Current plan is to move to a separate class for logging that handles the "are we in debug mode?" conditional, and standardise formatting. I'm hoping to start work on that shortly after this PR is accepted.

data/src/GameManager.js
michael-j-green: I could probably clean this up since there are multiple calls to the `this.EJS.downloadType` array which does stores the configuration of which items should be cached (eg cores and ROMs should be cached, but user content like save states should not be (fall back to current behaviour)).

gantoine: i don't think that's necessary, it was more on how the params are passed:

```
await this.EJS.downloadFile(url, {
  type: this.EJS.downloadType.support.name,
  method: "GET", 
  headers: {},
  body: null,
  onProgress: null, 
  onComplete: null,
  timeout: 30000,
  responseType: "arraybuffer",
  forceExtract: false,
  dontCache: this.EJS.downloadType.support.dontCache
});
```

then you could simplify to:

```
await this.EJS.downloadFile(url, {
  type: this.EJS.downloadType.support.name,
  dontCache: this.EJS.downloadType.support.dontCache
});
```

but don't make the change now, this PR is already sizable enough!

gantoine (approved):

data/src/GameManager.js
michael-j-green: This is a good point, and might make a good candidate for later refactoring.

You're certainly right about the size of it :D It's working now, and i don't want to change anything lest I break something haha

S-Coding23:
> hmm is there someone else to test this

I am trying to test this now, but I’m seeing this error in the browser console:
```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'enabled')
    at EmulatorJS.createBottomMenuBar (emulator.js:2194:37)
    at EmulatorJS.bindListeners (emulator.js:1296:14)
    at new EmulatorJS (emulator.js:340:14)
    at loader.js:189:27
```

michael-j-green:
> > hmm is there someone else to test this
> 
> I am trying to test this now, but I’m seeing this error in the browser console:
> 
> ```
> Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'enabled')
>     at EmulatorJS.createBottomMenuBar (emulator.js:2194:37)
>     at EmulatorJS.bindListeners (emulator.js:1296:14)
>     at new EmulatorJS (emulator.js:340:14)
>     at loader.js:189:27
> ```

Thanks for that, I'll take a look at that today.

michael-j-green:
> > hmm is there someone else to test this
> 
> I am trying to test this now, but I’m seeing this error in the browser console:
> 
> ```
> Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'enabled')
>     at EmulatorJS.createBottomMenuBar (emulator.js:2194:37)
>     at EmulatorJS.bindListeners (emulator.js:1296:14)
>     at new EmulatorJS (emulator.js:340:14)
>     at loader.js:189:27
> ```

This should hopefully now be resolved.

S-Coding23:
This is really amazing. I have only tested psx, but with .chd and .zip. and both are blazing fast! Not sure how big of a deal it is, but I do get this back in the logs with both formats:
```
Error reading directory: /proc/self ErrnoError {name: 'ErrnoError', errno: 63}
listDir | @ | GameManager.js:481
-- | -- | --
  | listDir | @ | GameManager.js:475
  | listDir | @ | GameManager.js:475
  | startGame | @ | emulator.js:1261
  | startGameFromDownload | @ | emulator.js:1140
  | (anonymous) | @ | emulator.js:1157
  | await in (anonymous) |   |  
  | downloadFiles | @ | emulator.js:1158
  | (anonymous) | @ | emulator.js:1202
  | Promise.then |   |  
  | initModule | @ | emulator.js:1200
  | (anonymous) | @ | emulator.js:819
  | script |   |  
  | createElement | @ | emulator.js:76
  | initGameCore | @ | emulator.js:816
  | processCore | @ | emulator.js:748
  | gotCore | @ | emulator.js:696
  | (anonymous) | @ | emulator.js:812
  | Promise.then |   |  
  | downloadGameCore | @ | emulator.js:758
  | startButtonClicked | @ | emulator.js:626
```

michael-j-green:
> This is really amazing. I have only tested psx, but with .chd and .zip. and both are blazing fast! Not sure how big of a deal it is, but I do get this back in the logs with both formats:
> 
> ```
> Error reading directory: /proc/self ErrnoError {name: 'ErrnoError', errno: 63}
> listDir | @ | GameManager.js:481
> -- | -- | --
>   | listDir | @ | GameManager.js:475
>   | listDir | @ | GameManager.js:475
>   | startGame | @ | emulator.js:1261
>   | startGameFromDownload | @ | emulator.js:1140
>   | (anonymous) | @ | emulator.js:1157
>   | await in (anonymous) |   |  
>   | downloadFiles | @ | emulator.js:1158
>   | (anonymous) | @ | emulator.js:1202
>   | Promise.then |   |  
>   | initModule | @ | emulator.js:1200
>   | (anonymous) | @ | emulator.js:819
>   | script |   |  
>   | createElement | @ | emulator.js:76
>   | initGameCore | @ | emulator.js:816
>   | processCore | @ | emulator.js:748
>   | gotCore | @ | emulator.js:696
>   | (anonymous) | @ | emulator.js:812
>   | Promise.then |   |  
>   | downloadGameCore | @ | emulator.js:758
>   | startButtonClicked | @ | emulator.js:626
> ```

Awesome! Thanks for the feedback!

`listDir` is only really called while in debug mode and is intended to dump a recursive directory tree for debugging purposes - it's meant as a quality of life thing for devs when trying to locate files in the virtual file system.

At present it does generate an error when trying to list the contents of directories like `/proc`. It's expected at this point in the implementation, though I might put a filter in place to make it skip system directories like this.

Edit: fixed. System directories `/dev`, `/proc`, and `/sys` are now skipped when running `listDir();`

ethanaobrien (approved):

ethanaobrien:
I hope nothing breaks!

michael-j-green:
> I hope nothing breaks!

I was thinking perhaps we should release the next version as a beta release, and call out major changes like this as potentially introducing new bugs.

I'm certain it's solid, but I'm sure there's going to be an edge case out there somewhere that I won't have thought about during testing.

Issue #1103, opened 2025-10-08
Feature Request: beforeunload event listener update


duanyu5871:
Hello, 
Recently, when I tried to add a "beforeunload" event listener to my webpage to display a confirmation button when exiting, I found that although I could click the cancel button to prevent leaving the page, the emulator had already stopped working. So I checked the source code and found this in emulator.js file. 
```
this.addEventListener(window, "beforeunload", (e) => {
    if (!this.started) return;
    this.callEvent("exit");
});

this.EJS.on("exit", (()=>{
    this.EJS.failedToStart || (this.saveSaveFiles(),
    this.functions.restart(),
    this.saveSaveFiles()),
    this.toggleMainLoop(0),
    this.FS.unmount("/data/saves"),
    setTimeout((()=>{
        try {
            this.Module.abort()
        } catch (t) {
            console.warn(t)
        }
    }
    ), 1e3)
}
```
When the beforeunload listener is triggered, both methods have run, which causes this problem. I can't solve this problem unless modify emulator.js file.

Is it possible to pull a request to control whether the window pops up through the setting like window.EJS_something with default "false", so as to avoid the pop-up for those who don't want to use this function?

ethanaobrien:
I think that's reasonable, I can add the `EJS_askBeforeUnload` boolean when I have the time, or you can PR.

duanyu5871:
Sorry, I spent two nights trying , but still not able to solve all problems.  My codes tried to show a confirmation button is like this
```
this.addEventListener(window, "beforeunload", (e) => {
    if (window.EJS_askBeforeUnload) {
        e.preventDefault(); 
        e.returnValue = '';
    } 
    ......
});
```
However, it seems that no methods to determine whether the Confirm button or the Cancel button has been clicked in this default dialogue box, which means there is no way to set whether to trigger the "exit" event for different buttons. 

Although I have been following this project for a long time, I am quite an amateur in programming. Given my current skill level, I may not be able to submit a PR that ensures it runs logically.

ethanaobrien:
I believe all youre missing is a `return` inside that if statement :)

duanyu5871:
Thank you so much for your help! The RP has been merged, and I have also learned about the beforeunload event as well as gained a better understanding of code style.

Additionally, I will continue to follow up on and learn about the visibilitychange event mentioned by michael-j-green.

Thank you again!

PR #1107, opened 2025-10-12
beforeunload event listener update

duanyu5871:
[issues1103](https://github.com/EmulatorJS/EmulatorJS/issues/1103) 
Thank you for your suggestion. I've tried modifying this solution, and at least it can run basically. However, there are still some flaws. After clicking confirm in the pop-up window, it's still unable to make `this.callEvent("exit")` work as expected. I've also tried using the unload event listener, but it still can't ensure that `callEvent("exit")` runs stably.

But at the very least, it doesn't alter the original operational logic. Meanwhile, when EJS_askBeforeUnload == true, it prevents accidental exits caused by made a wrong press.

If you have a better solution, I will very happy to learn from it.   :)

Thank you again

### DIFF
--- a/data/loader.js
+++ b/data/loader.js
@@ -127,6 +127,7 @@
     config.browserMode = window.EJS_browserMode;
     config.shaders = Object.assign({}, window.EJS_SHADERS, window.EJS_shaders ? window.EJS_shaders : {});
     config.fixedSaveInterval = window.EJS_fixedSaveInterval;
+    config.disableAutoUnload = window.EJS_disableAutoUnload;
 
     let systemLang;
     try {
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -1159,9 +1159,9 @@ class EmulatorJS {
             }, 0);
         });
         this.addEventListener(window, "beforeunload", (e) => {
-            if (window.EJS_askBeforeUnload) {
+            if (this.config.disableAutoUnload) {
                 e.preventDefault();
-                e.returnValue = '';
+                e.returnValue = "";
                 return
             } 
             if (!this.started) return;
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -1159,6 +1159,11 @@ class EmulatorJS {
             }, 0);
         });
         this.addEventListener(window, "beforeunload", (e) => {
+            if (window.EJS_askBeforeUnload) {
+                e.preventDefault();
+                e.returnValue = '';
+                return
+            } 
             if (!this.started) return;
             this.callEvent("exit");
         });
--- a/index.html
+++ b/index.html
@@ -259,7 +259,6 @@
                 window.EJS_DEBUG_XX = enableDebug;
                 window.EJS_disableDatabases = true;
                 window.EJS_threads = enableThreads;
-                window.EJS_askBeforeUnload = false;
                 if (browserMode) {
                     window.EJS_browserMode = browserMode;
                 }
--- a/index.html
+++ b/index.html
@@ -259,6 +259,7 @@
                 window.EJS_DEBUG_XX = enableDebug;
                 window.EJS_disableDatabases = true;
                 window.EJS_threads = enableThreads;
+                window.EJS_askBeforeUnload = false;
                 if (browserMode) {
                     window.EJS_browserMode = browserMode;
                 }


ethanaobrien:
I've cleaned this up and renamed it to `EJS_disableAutoUnload` to make limitations known. AFAIK it's not possible to know if the user closes the page because.. well it will be closed.

ethanaobrien (approved):

michael-j-green:
Just reading the documentation for `beforeunload` (https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event), and it's recommending that you should listen for the `visibilitychange` event if the goal is to save state (which I think of the majority of users is probably the case). Documentation here: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event with supporting documentation here: https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/

Possible use cases:
- `beforeunload`
  - when closing a tab, this can be used to display a "are you sure you want to close?" alert. However, mobile browsers do not fire this event
- visibilitychange`
  - can be used to kick off save events when a tab moves to the background - which on mobile would be invaluable since if the user moves to another app, the host OS may kill the browser session unexpectedly to free up RAM for another task (iOS is aggressive in this, but Android does this too in low RAM situations). Situations where on mobile the browser moves to the background include:
    - user app switching
    - an incomming call
    - accidentally going to the home screen (with on screen controls, it can be easy to trigger going to the home screen)
  - can be used to automatically pause and unpause the emulation in progress - this could be a good UX for the user as (notably on Apple devices), emulation can become choppy in the background which will produce nasty audio

It should be noted that `visibilitychange` is supported by all modern browsers, while `beforeunload` is spotty at best.

`beforeunload`:
image

`visibilitychange`:
image

Should we try and support both events? Thinking EJS could be the active listener for those events, and then the hosting page can attach a callback to those.

ethanaobrien:
> can be used to kick off save events when a tab moves to the background

If save events involve a user prompt is this relevant to us?

michael-j-green:
> > can be used to kick off save events when a tab moves to the background
> 
> If save events involve a user prompt is this relevant to us?

A save event could be handled silently by triggering a save state that the surrounding page can capture and send to a host somewhere or to localstorage for retrieval later. Same with srm saves.

A dev could use `visibilitychange` to capture when the page is being moved to the background or closed to trigger a silent save and pause the emulator (in the event of the page just being backgrounded); and then when the page is restored, the emulator could be setup to basically pick up where it left off. This would have the experience of a native game where you can leave the game for a bit, do something else, and then return to where you left off.

I personally wouldn't emit an alert asking if the user wants to quit - because I like the above auto save option. I do like having the option available for others though. And if there is a core that doesn't create srm's or doesn't play nicely with save states, the dev could then send the alert to warn the user that closing will be bad.

But from my reading of the docs, anything we tie to `beforeunload` is likely not going to be consistently executed by the browser - hence my recommendation for providing the scaffolding for `visibilitychange`.

ethanaobrien:
My concern is that I don't believe SRMs are universally stable enough across all cores and browsers that I want to rely on system saves for automatically saving. I'm fine with adding another auto-save trigger/event using visibilitychange. It doesnt really make sense to automatically store save states, since how are we presenting this to the user? Are we mandating an auto-load save state when you start the game back up?

SRM saves also depend on the user having saved in game, which we cannot check. A dialog asking the user to confirm they have no changes to save makes more sense. It's opt-in anyways

michael-j-green:
I agree with you on the SRM issues. It's completely inconsistent. The saving on visibilitychange was just an example of what a dev could do with that hook. No mandating or anything. Just exposing the call back as a quality of life thing - as technically, both beforeunload and visibilitychange events could be captured by the page hosting EJS. EJS wouldn't do anything besides expose the event for the page dev to attach a callback.

That callback could then be responsible for showing dialogs, saving state, etc.

How I would implement using visibilitychange in Gaseous:
- user is playing and goes to the home screen
- this emits the visibilitychange event (there is an argument supplied that says the page is no longer visible). This event the triggers an auto-save of the state, and pauses the emulation, and store a flag in sessionstorage called isRunning to say a game is in progress (sessionstorage is cleared on browser reload (note this is distinct from refresh))
- during this time, depending on the OS, the page can be completely unloaded from memory forcing a refresh the next time it's displayed.
- user returns to the browser their game was in - this emits another visibilitychange event (the argument supplied says the page is visible again). Using this event I would check for the isRunning flag in sessionstorage - if present, just resume the emulation (because we know the page wasn't unloaded), if it's not present, load the state from storage, and resume.
- should the user navigate away from the page (forward, back, or href), and then back again, the game will reload where the user left off (as it's the same session). But clicking the start button would trigger a clean start clearing the state and isRunning flag.

This presents a good user experience to the user because their game state is preserved as much as possible, which on mobile can be a pain.

ethanaobrien:
Why would emulatorjs need to handle anything then? It's a document event, not anything special

michael-j-green:
Too be fair, that's true for this entire PR right? 😄 

I only mentioned providing those callbacks because that's what this PR was trying to do.

ethanaobrien:
No, this PR doesnt have anything to do with page visibility

Currently, when EmulatorJS gets the `beforeunload` event, it will flush indexeddb and exit. However, if the embedding page wants to prompt the user with the "Are you sure you want to exit" browser prompt (by calling `preventDefault`) and the user chooses to stay on the page, emulatorjs will have already quit. This PR adds the ability to tell emulatorjs to not exit when `beforeunload` is fired

michael-j-green:
I know 😄 

I brought up `visiblitychanged` because the documentation for `beforeunload` recommends not using `beforeunload` because it's not reliable.

Then we got just a little bit off track haha

ethanaobrien:
Are you saying you'd like to remove the exit on before unload entirely?

michael-j-green:
No... as there are probably people using it.

Though we should probably apply this PR, then move it to deprecated with some advice in the documentation around how to implement `beforeunload` at the page level with a thingie that says "the documentation for `beforeunload` recommends using `visibilitychanged` as it's more reliable".

ethanaobrien:
I don't see any reason to deprecate this. The exit event is not only triggerable by the beforeunload event. All beforeunload does is fire exit.

michael-j-green:
Fair enough 😄

michael-j-green (approved):

Issue #1039, opened 2025-07-03
Save state text is inconsistent (i think)


joethun:
This is on the en-US.json file:
```
"SAVE LOADED FROM BROWSER": "SAVE LOADED FROM BROWSER",
"SAVE SAVED TO BROWSER": "SAVE SAVED TO BROWSER",
```
I think this should be "SAVED STATE TO BROWSER" so it would be more consistent with the other save state texts? Also "SAVE SAVED" sounds weird.

ethanaobrien:
State = a copy of that exact moment in the game, to where you can instantly pick back up

Save = system storage, eg: SRM

joethun:
Yes, but this text appears when you have the state location set to browser and you click the save and load state buttons. They in fact load you to the exact point you saved at.

ethanaobrien:
You're clicking the save state buttons (not the save save buttons)?

joethun:
Yass

ethanaobrien:
Yep its wrong 👍

PR #1116, opened 2025-10-20
Fix Issue #1039 (Save state text is inconsistent)

joethun:
Fix Issue #1039

Google Translate was used for the translations, most of them translate to "Status loaded from browser" and "Status saved to browser"


ethanaobrien (approved):
Thanks for this

PR #1161, opened 2026-01-17
feat: controls autofire support

zees-dev:
### Summary

This PR adds configurable autofire support to EmulatorJS, allowing buttons to automatically repeat press/release cycles at a configurable interval.
Autofire can be configured per-button through the Control Settings UI or via the `EJS_defaultControls` API, and works across all input methods (keyboard, gamepad, and virtual gamepad).

### Features

- **Per-button autofire toggle** - Enable/disable autofire for any button via checkbox in Control Settings
- **Configurable intervals** - Global default interval with optional per-button overrides
- **Unified control system** - Single configuration applies to keyboard, physical gamepad, and virtual gamepad
- **Persistent settings** - Autofire preferences are saved to localStorage per-game
- **Analog axis exclusion** - Autofire is automatically disabled for analog stick axes (not applicable)

#### Global Default Interval
```javascript
EJS_defaultAutoFireInterval = 100;  // Default: 100ms
```

#### Per-Button Configuration
```javascript
EJS_defaultControls = {
  0: {  // Player 0
    8: { autofire: true },                    // A button - uses global interval
    0: { autofire: true, autoFireInterval: 50 }  // B button - custom 50ms interval
  }
}
```

### Control Settings UI

Added an "Autofire" column to the Control Settings menu with checkboxes for each button (excluding analog axes).

Screenshot 2026-01-17 at 4 01 13 PM

#### Files Modified

| File | Changes |
|------|---------|
| `data/loader.js` | Added `EJS_defaultAutoFireInterval` config |
| `data/src/emulator.js` | Core autofire implementation |
| `data/localization/*.json` | Added "Autofire" translation key (19 files) |

#### Key Implementation Points

- **`this.analogAxes`** - Centralized definition of analog stick indices `[16-23]` that don't support autofire
- **`this.autofireIntervals`** - Tracks active autofire intervals by `"player-button"` key
- **Helper methods**: `startAutofire()`, `stopAutofire()`, `stopAllAutofire()`, `isAutofireEnabled()`, `getAutofireInterval()`

#### Autofire Cleanup Triggers

Autofire is automatically stopped when:
- Window loses focus (blur event)
- Emulator is paused
- Control settings are Reset or Cleared
- Button is released
- Autofire checkbox is unchecked

### Breaking Changes

None. This is a purely additive feature.

## Use of AI

This PR was primarily developed using claude 4.5 Opus (2 iterations - plan first; multiple prompts).
The functionality was tested manually on browser and mobile device.

### DIFF
--- a/data/loader.js
+++ b/data/loader.js
@@ -129,6 +129,7 @@
     config.fixedSaveInterval = window.EJS_fixedSaveInterval;
     config.disableAutoUnload = window.EJS_disableAutoUnload;
     config.disableBatchBootup = window.EJS_disableBatchBootup;
+    config.defaultAutoFireInterval = window.EJS_defaultAutoFireInterval;
 
     let systemLang;
     try {
--- a/data/loader.js
+++ b/data/loader.js
@@ -129,7 +129,6 @@
     config.fixedSaveInterval = window.EJS_fixedSaveInterval;
     config.disableAutoUnload = window.EJS_disableAutoUnload;
     config.disableBatchBootup = window.EJS_disableBatchBootup;
-    config.defaultAutoFireInterval = window.EJS_defaultAutoFireInterval;
 
     let systemLang;
     try {
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -3620,18 +3620,13 @@ class EmulatorJS {
 
                 if (["buttonup", "buttondown"].includes(e.type) && (controlValue === e.label || controlValue === e.index)) {
                     const inputValue = isAnalog ? 0x7fff : 1;
-                    if (e.type === "buttonup") {
-                        if (this.isAutofireEnabled(i, j) && !isAnalog) {
-                            this.stopAutofire(i, j);
-                        } else {
-                            this.gameManager.simulateInput(i, j, 0);
-                        }
+                    const isButtonUp = e.type === "buttonup";
+                    const value = isButtonUp ? 0 : inputValue;
+
+                    if (this.isAutofireEnabled(i, j) && !isAnalog) {
+                        isButtonUp ? this.stopAutofire(i, j) : this.startAutofire(i, j, inputValue);
                     } else {
-                        if (this.isAutofireEnabled(i, j) && !isAnalog) {
-                            this.startAutofire(i, j, inputValue);
-                        } else {
-                            this.gameManager.simulateInput(i, j, inputValue);
-                        }
+                        this.gameManager.simulateInput(i, j, value);
                     }
                 } else if (e.type === "axischanged") {
                     if (typeof controlValue === "string" && controlValue.split(":")[0] === e.axis) {
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -4052,18 +4052,14 @@ class EmulatorJS {
                     if (e.type === "touchend" || e.type === "touchcancel") {
                         e.target.classList.remove("ejs_virtualGamepad_button_down");
                         window.setTimeout(() => {
-                            if (this.isAutofireEnabled(0, value) && !isAnalog) {
-                                this.stopAutofire(0, value);
-                            } else {
-                                this.gameManager.simulateInput(0, value, 0);
-                            }
+                            this.stopAutofire(0, value);
+                            this.gameManager.simulateInput(0, value, 0);
                         })
                     } else {
                         e.target.classList.add("ejs_virtualGamepad_button_down");
+                        this.gameManager.simulateInput(0, value, downValue);
                         if (this.isAutofireEnabled(0, value) && !isAnalog) {
                             this.startAutofire(0, value, downValue);
-                        } else {
-                            this.gameManager.simulateInput(0, value, downValue);
                         }
                     }
                 })
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -3507,7 +3507,8 @@ class EmulatorJS {
         if (control && typeof control.autoFireInterval === "number") {
             return control.autoFireInterval;
         }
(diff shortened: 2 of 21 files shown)


data/src/emulator.js
```diff
@@ -238,7 +237,23 @@ class EmulatorJS {
...
+                }
+                for (const button in this.config.defaultControllers[player]) {
+                    this.defaultControllers[player][button] = Object.assign(
+                        {},
+                        this.defaultControllers[player][button] || {},
+                        this.config.defaultControllers[player][button]
+                    );
+                }
+            }
+        }
```
ethanaobrien, line 252:
```
if (this.config.defaultControllers) {
    for (const [player, buttons] of Object.entries(this.config.defaultControllers)) {
        this.defaultControllers[player] = this.defaultControllers[player] || {};

        for (const [button, config] of Object.entries(buttons)) {
            this.defaultControllers[player][button] = {
                ...(this.defaultControllers[player][button] || {}),
                ...config
            };
        }
    }
}
```

We can probably use the spread syntax here, something like this I believe should work

data/src/emulator.js
```diff
@@ -3452,11 +3566,24 @@ class EmulatorJS {
...
+                        } else {
+                            this.gameManager.simulateInput(i, j, 0);
+                        }
+                    } else {
+                        if (this.isAutofireEnabled(i, j) && !isAnalog) {
+                            this.startAutofire(i, j, inputValue);
+                        } else {
+                            this.gameManager.simulateInput(i, j, inputValue);
+                        }
+                    }
```
ethanaobrien: These if else blocks are very similar, it should be possible to combine them

data/loader.js
```diff
@@ -129,6 +129,7 @@
     config.fixedSaveInterval = window.EJS_fixedSaveInterval;
     config.disableAutoUnload = window.EJS_disableAutoUnload;
     config.disableBatchBootup = window.EJS_disableBatchBootup;
+    config.defaultAutoFireInterval = window.EJS_defaultAutoFireInterval;
```
ethanaobrien: Instead of adding a new config var for this, I think it'd be better to add this to the settings menu to allow users to configure this and allow setting via defaultOptions

data/src/emulator.js
```diff
@@ -3915,14 +4055,23 @@ class EmulatorJS {
...
+                            }
                         })
                     } else {
                         e.target.classList.add("ejs_virtualGamepad_button_down");
-                        this.gameManager.simulateInput(0, value, downValue);
+                        if (this.isAutofireEnabled(0, value) && !isAnalog) {
+                            this.startAutofire(0, value, downValue);
+                        } else {
+                            this.gameManager.simulateInput(0, value, downValue);
+                        }
```
ethanaobrien: I think this function should be the one to send the initial up/down input. Just handle it normally here and then do autofire checks

data/src/emulator.js
```diff
@@ -1962,6 +1977,9 @@ class EmulatorJS {
                 }
             }
             this.gameManager.toggleMainLoop(this.paused ? 0 : 1);
+            if (this.paused) {
+                this.stopAllAutofire();
+            }
```
ethanaobrien: I don't like the idea of having to place `stopAllAutofire` everywhere that may need it. We can probably check inside the settimeout statement if the emulation is still running.

If the user really opens a dialog or switches windows *while* holding the button, they can easily just press the key again. The keydown/up implementation doesnt account for this because I don't think it's worth the effort to account for every possible "issue"

ethanaobrien (changes requested):

data/src/emulator.js
zees-dev: https://github.com/EmulatorJS/EmulatorJS/pull/1161/commits/e90f286abda18a5ac041446a4a10a97440626825

data/src/emulator.js
zees-dev: https://github.com/EmulatorJS/EmulatorJS/pull/1161/commits/b3d486c7d1d0424b7376be0891d1bf2eb3d3273d

data/src/emulator.js
zees-dev: https://github.com/EmulatorJS/EmulatorJS/pull/1161/changes/90fec707f16af7d34469c831f71ece8dd5a2db4b

data/loader.js
zees-dev: https://github.com/EmulatorJS/EmulatorJS/pull/1161/changes/004ec352523c53d937686f037e45071af52fb81d

data/src/emulator.js
zees-dev: https://github.com/EmulatorJS/EmulatorJS/commit/91dcb94e51e86d6615175a921560f86d113dc260

michael-j-green: > [91dcb94](https://github.com/EmulatorJS/EmulatorJS/commit/91dcb94e51e86d6615175a921560f86d113dc260)

Please provide a brief description of the change you made in the commit in addition to the link so that we can check that your intentions match the code as implemented.

data/src/emulator.js
michael-j-green: > [90fec70](https://github.com/EmulatorJS/EmulatorJS/commit/90fec707f16af7d34469c831f71ece8dd5a2db4b)

Again, please provide some description to the change implemented in the commit (other than the commit message).

ethanaobrien (approved):
I think this looks good, ~~Could you screenshot UI changes?~~

Oh, it was in your original post. I'd like to take a look at it and see if it can be a little cleaner

ethanaobrien:
Actually I'll merge and fix it later

PR #1171, opened 2026-02-22
Update Turkish Translation

Notenlish:
Hello, I noticed that there were a couple of weirdly worded translations in the translations for Turkish language, so I thought I would fix them.


Notenlish:
It appears that the zip I downloaded from the latest release had a different key value pair for `"LOADED STATE FROM BROWSER"` and `"SAVED STATE TO BROWSER"`, fixed that now

ethanaobrien (approved):
Thank you!

PR #1172, opened 2026-02-24
chore: adding khmer support translation

Novsochetra:
### After Change

https://github.com/user-attachments/assets/873a0310-57c4-439d-9066-a2a5bdb51196


ethanaobrien (approved):
Thank you very much

PR #1192, opened 2026-03-31
Update index.html

israelreyes38575:
Add Wii & GameCube


ethanaobrien:
💀

duanyu5871:
@israelreyes38575 Hello, I don't quite understand what this means. From what I remember, the Wii and GameCube are not supported yet, and there is no corresponding support in the core either. Besides, iso appears to be the extension name for disc image files, which doesn't seem to be directly related to game consoles. I'm not sure what you're trying to do here.

PR #1198, opened 2026-04-06
Fixed opposing axis value being retained if not passing trough deadzone

T-Thurow:
**Summary**
This PR fixes a bug where an axis value is retained when it "jumps" directly between opposing maximums (e.g., -1 to 1) without passing through the deadzone (0).

**Problem**
This can occur with controllers that simulate analog sticks via a D-Pad (like the 8BitDo SF/C 30). When a user switches directions instantly (e.g., Left to Right), the input value might jump from -1 directly to 1.
Because the value never hits 0, the logic fails to reset the previous direction's state. This results in both opposing axis values (e.g., LEFT_STICK_X:+1 and LEFT_STICK_X:-1) being active simultaneously, which is physically impossible for a standard stick.

**Fix**
Added a check to detect these direct transitions. If the value jumps from one extreme to the other, the opposite direction is now manually forced to 0.

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -3786,13 +3786,19 @@ class EmulatorJS {
         if (gamepadIndex < 0) {
             return; // Gamepad not set anywhere
         }
-        const value = function (value) {
-            if (value > 0.5 || value < -0.5) {
-                return (value > 0) ? 1 : -1;
-            } else {
-                return 0;
-            }
-        }(e.value || 0);
+
+        const toIntValue = (value) => {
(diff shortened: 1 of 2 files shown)


ethanaobrien (commented):
The standard spacing for this project is 4 spaces (rather than 2). Could you apply this change before this is merged

T-Thurow:
> The standard spacing for this project is 4 spaces (rather than 2). Could you apply this change before this is merged

Now the indention should be correct.

ethanaobrien (approved):
Thank you for the fix!

Issue #784, opened 2024-02-23
Any plans for supporting Webpack/TypeScript


ruochenjia:
Currently the modules published on `npm` is not importable from a Webpack module, which can be very difficult to integrate with projects that uses webpack. Furthermore, the module also doesn't seem to have a TypeScript declaration (`.d.ts`) file, which makes it impossible to be imported to a TypeScript project without casting to `any`. (casting to `any` would require the user to check the documentations in JS for the corresponding API, and it is not checked while compiling, which loses the benefit of using TypeScript).

If this is supported, it would be much easier for other projects to use it as an external library.

lnfel:
Hi just been looking for the same thing. apparently the author has yet to learn how to export modules. And they had just recently learned how to publish to npm as seen on #755. Enabling proper exports for library usage would take quite a time. I can look into making a PR but my fingers are crossed atm.

ethanaobrien:
> apparently the author has yet to learn how to export modules.

I don’t understand what modules that you’re needing to be exported, or how you would use the exported modules. (Same reason it wasn’t published to npm for the longest time, I don’t understand how it would be used)

One other issue is I am also attempting to maintain support for older browser versions, older than when module importing/exporting existed, while trying to not have any duplicate loader files. 

I don’t plan to add a typescript declaration file (I try to avoid typescript, I don’t care for type checking when I already know what the type is), but if anyone else wants to, please do submit a pr

williamisnotdefined:
@ethanaobrien would a typescript pr be really welcome or would it bring more headache? I would be happy to contribute for your project. Feel free to point some other direction if TS is not a good fit for the moment, although I think community would like it.

ethanaobrien:
@williamisnotdefined I would definitely accept a PR with typescript definitions, though I don't want to re-write the project to use typescript as I don't want to need a "build" step before running the browser

PR #1199, opened 2026-04-08
Add TypeScript definitions

williamisnotdefined:
Added TypeScript definitions
Discussion: https://github.com/EmulatorJS/EmulatorJS/issues/784

### DIFF
--- a/.github/workflows/eslint.yml
+++ b/.github/workflows/eslint.yml
@@ -18,9 +18,10 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: '20'
+        cache: npm
         
     - name: Install dependencies
-      run: npm install
+      run: npm ci --omit=optional
       
     - name: Run ESLint
       run: npx eslint .
--- a/.github/workflows/types.yml
+++ b/.github/workflows/types.yml
@@ -1,0 +1,30 @@
+name: Type Definitions
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+jobs:
+  check-types:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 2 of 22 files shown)


williamisnotdefined:
After 3 commits now I think it is ready to review.

williamisnotdefined:
@ethanaobrien if you'd like I could create a github action to run after each PR the check:types, which is kind of a "smoke test" for types.

ethanaobrien:
>if you'd like I could create a github action to run after each PR the check:types, which is kind of a "smoke test" for types.

Yes please, this would be helpful

williamisnotdefined:
- **`npm ci` instead of `npm install`**
  - Installs exactly what is defined in `package-lock.json`
  - Makes CI more reproducible
  - Fails if `package.json` and `package-lock.json` are out of sync
  - Better suited for CI environments

- **`cache: npm`**
  - Reuses downloaded packages between GitHub Actions runs
  - Speeds up dependency installation

- **`--omit=optional`**
  - Skips optional dependencies that are not required for linting (e.g. `@emulatorjs/cores`)
  - Makes the job lighter and less prone to failures due to unnecessary downloads

- **Consistency**
  - Aligns the ESLint workflow with the new type-check workflow

### Tradeoff

- CI becomes stricter, since it will fail if the lockfile is outdated  
  → This is generally a **good thing**, as it enforces consistency

### Summary

- Faster CI  
- More predictable CI  
- More reproducible installs  
- No impact on linting results

williamisnotdefined:
In my company we are used to work with [husky](https://typicode.github.io/husky/).
What about adding it to the project?
we could automatically run type:check before a git push, aborting it if fails.

ethanaobrien:
I'm not sure how I feel about merging ai generated code for a stack I'm not familiar with myself.

PR #1205, opened 2026-04-19
Add exit confirmation prompt and exit callback functionality

michael-j-green:
Introduce an option to suppress the confirmation prompt before exiting the emulator and implement a callback function that triggers upon exit.

Adds options:
- `EJS_askBeforeExit` (boolean - default true) determines if the "Are you sure you want to exit?" popup appears
- `EJS_onExit` (callback) adds a callback function to execute on exit

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -2387,7 +2387,7 @@ class EmulatorJS {
         let exitMenuIsOpen = false;
         const exitEmulation = addButton(this.config.buttonOpts.exitEmulation, async () => {
             if (exitMenuIsOpen) return;
-            if (this.config.askBeforeExit === true || this.config.askBeforeExit === undefined) {
+            if (this.config.askBeforeExit !== false) {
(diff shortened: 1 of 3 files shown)


data/src/emulator.js
```diff
@@ -2387,56 +2387,65 @@ class EmulatorJS {
...
-            close.classList.add("ejs_cheat_close");
-            header.appendChild(title);
-            header.appendChild(close);
-            popup.appendChild(header);
-            this.addEventListener(close, "click", (e) => {
-                exitMenuIsOpen = false
-                popups[0].remove();
-            })
-            popup.appendChild(this.createElement("br"));
+            if (this.config.askBeforeExit === true || this.config.askBeforeExit === undefined) {
```
ethanaobrien: This should be `!== false`

michael-j-green: Fixed! 😀

ethanaobrien (approved):

PR #1232, opened 2026-07-06
Fix screen capture filename month

yesidodo73:
This fixes generated screenshot and screen recording filenames using a zero-based month value from Date#getMonth().

### DIFF
--- a/data/src/emulator.js
+++ b/data/src/emulator.js
@@ -768,6 +768,9 @@ class EmulatorJS {
         parts.splice(parts.length - 1, 1);
         return parts.join(".");
     }
+    getFileNameTimestamp(date) {
+        return date.toISOString().replace(/[-:]/g, "").replace(/\.\d{3}Z$/, "Z");
+    }
     saveInBrowserSupported() {
         return !!window.indexedDB && (typeof this.config.gameName === "string" || !this.config.gameUrl.startsWith("blob:"));
     }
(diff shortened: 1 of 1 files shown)


michael-j-green:
Hi @yesidodo73; thanks for your contribution!

While I'm glad to see dates normalised to ensure they always contain the same number of characters (the term you're looking for is leading-zero, not zero based), I don't think doubling down on the US style of dates is appropriate.

When using dates as strings, we should try to standardise on international standards that leave no ambiguity such as ISO 8601 which calls for a string with the format  `yyyyMMddThh:mm:ss`, with the encoded date itself being in UTC to ensure that time zones and daylight savings time can be computed correctly.

These style of dates also sort properly when sorted by string.

@ethanaobrien; what do you reckon?

yesidodo73:
> Hi @yesidodo73; thanks for your contribution!
> 
> While I'm glad to see dates normalised to ensure they always contain the same number of characters (the term you're looking for is leading-zero, not zero based), I don't think doubling down on the US style of dates is appropriate.
> 
> When using dates as strings, we should try to standardise on international standards that leave no ambiguity such as ISO 8601 which calls for a string with the format `yyyyMMddThh:mm:ss`, with the encoded date itself being in UTC to ensure that time zones and daylight savings time can be computed correctly.
> 
> These style of dates also sort properly when sorted by string.
> 
> @ethanaobrien; what do you reckon?

Thanks for the feedback!

I’ve updated the screenshot and recording filenames to use a filename-safe ISO 8601 style UTC timestamp instead of the previous US style date format.

ethanaobrien (commented):
This looks good, thanks

PR #1233, opened 2026-07-06
Fix screen recording menu state

yesidodo73:
Only switch the menu to the stop-recording state after screenRecord() returns a recorder.

Also restore the start/stop buttons when the recorder stops or errors.


ethanaobrien (approved):
Thanks for the fix

PR #1242, opened 2026-07-25
Enable Core Options v2

RobLoach:
Currently, EmulatorJS uses Core Options v1, which means that there is just a key, and a value for each variable. This updates it to allow use of Core Options V2, which brings in human readable labels, descriptions, etc.

Enabled by https://github.com/EmulatorJS/RetroArch/pull/45 .

### Before

Screenshot from 2026-07-25 15-39-09

### After

Screenshot from 2026-07-25 15-39-34


ethanaobrien (approved):
Thanks, this is nice

RobLoach:
Sidenote: Core Options v2 also provides Categories, and I tried it out, and found the UI ended up being a bit more complicated because a lot of it was split into Audio/Video/Gameplay/etc sub-menus...

It could be something we re-consider though. Maybe use sub-sections instead of sub-menus. I'll leave it for now though. At the very least, this is a lot more user-friendly.

Thanks a lot for the review and merge.

PR #1250, opened 2026-08-13
docs: update star history links to star-history.dera.page

OctoBored:
The current star history chart on README is broken because it relies on GitHub stargazers API which has been restricted for this purpose. This PR switches the star history widget to an alternative data source (star-history.dera.page) that requires no API token and serves both the API and the frontend from the same domain.

Only the star history section was modified; all other sections remain untouched. No badge endpoints exist in this project.


ethanaobrien (approved):
Thank you for the fix

Issue #1256, opened 2026-08-25
ColecoVision/gearcoleco Control Scheme and input bindings are broken


Olliebrown:
## Observed Behavior
- The inputs '0' and '9' on the keypad are missing from the mappable control scheme.
- The labels for available keypad keys in the control scheme do NOT match the actual inputs that fire.
- Generally, all even numbered inputs are swapped with their odd numbered counterparts.
- When on mobile, the Virtual Gamepad omits all the keypad buttons making most games unplayable without an external controller.

## Expected Behavior
- gearcoleco properly exposes all these inputs in retroarch and they can be mapped.
- They should be presented in the control scheme and be mappable/re-mappable like any other inputs.
- They should map to the proper internal inputs that match their labels.
- The virtual gamepad should present these buttons as they are critical to playing almost all colecovision games.

## Discussion
The ColecoVision (and inteliVision) have a unique controller problem in that they have a large keypad with 12 buttons on it (0 - 9, # and *).  MANY games uses these button for special functions and while there is some bias towards the "#" and "*" keys being more important, no one of these keys can be ignored or left out if you want to support games.

While historically I see a lot of posts stating that it is impossible to support all these buttons, I did some investigationg and concluded this is simply not true (anymore).  They have very strange index mappings, but they are all there and they work if you setup the control scheme properly.

I have found a way to fix this in the setup of the default control scheme.  changes within the "createControlSettingMenu()" which are isolated to only happen when the core is "coleco" allow exposing all the proper inputs.  Similar updates inside "setVirtualGamepad()" which are also isolated to the "coleco" core can add all the missing keypad buttons.

I have built and tested this fix with the "Final Test" game rom which allows for testing controller inputs and the fix is well isolated and properly exposes all these controls.  I will offer this fix as a pull request.

I will note that the default keyboard mappings and gamepad mappings are still very bad and need to be updated for this rom, but I want to focus on the game scheme exposing them so that the bad mappings CAN be fixed first.  I will investigate offering a better mapping separately (that one is much more subjective).  I fell the missing and scrambled control scheme is not subjective, it's just a bug that needs fixing!

PR #1257, opened 2026-08-25
Bugfix/gearcoleco input mapping (#1256)

Olliebrown:
This is a fix for the gearcoleco control scheme mapping and virtual gamepad (Issue #1256).  All low level inputs are properly available from gearcoleco and this fix exposes them all (including '0' and '9') and it fixes mappings so they match their visual labels.

It also updated the virtual gamepad with keypad buttons in the lower corners.  Almost all games (all that I know of) require these keys in some form to be playable so this update makes that possible on mobile platforms without having to connect an external gamepad.


ethanaobrien (approved):
Looks good to me

Issue #1191, opened 2026-03-27
[Bug] Game lags when save to browser is used (NIGHTLY)


joethun:
Games seem to lag when the save state to browser button is pressed. This is especially annoying with an auto-save state system.  I have noticed this on PSX and DS.

Issue #1074, opened 2025-08-31
[Bug] Mobile Context Menu Appears On non-touch devices


joethun:
<img width="1365" height="580" alt="Image" src="https://github.com/user-attachments/assets/4fa36ca8-c94b-4b57-9ede-b178fdbe9b9f" />
I've noticed this on 3 separate 12 inch Chromebooks, 2 of which had touchscreens.

ethanaobrien:
@allancoding I've also noticed my PC is now detected as touch

joethun:
My 16 inch touchscreen laptop doesn't have this issue 🤔

ethanaobrien:
The touchscreen laptop doesn't have the hamburger menu? It should

joethun:
It has the menu it's just in the bottom menu bar

ethanaobrien:
That's a separate menu. See your original screenshot

joethun:
So it should have the hamburger menu in the top right?

ethanaobrien:
Yes. A touchscreen should have the menu in the top right and a not touchscreen should not

allancoding:
I blame browsers. (I'll look into it)

PR #1264, opened 2026-09-06
Fix audio muting and slider not working, save state stutter, and mobile context menu appearing on desktops.

joethun:
This fixes issues #1191 and #1074 and also an issue where the audio controls don't work.

### DIFF
--- a/data/src/GameManager.js
+++ b/data/src/GameManager.js
@@ -5,7 +5,7 @@ class EJS_GameManager {
         this.FS = this.Module.FS;
         this.functions = {
             restart: this.Module.cwrap("system_restart", "", []),
-            //saveStateInfo: this.Module.cwrap("save_state_info", "string", []),
+            saveStateInfo: this.Module.cwrap("save_state_info", "number", []),
             loadState: this.Module.cwrap("load_state", "number", ["string", "number"]),
             screenshot: this.Module.cwrap("cmd_take_screenshot", "", []),
             simulateInput: this.Module.cwrap("simulate_input", "null", ["number", "number", "number"]),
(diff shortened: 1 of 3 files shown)


data/src/GameManager.js
```diff
@@ -205,10 +205,19 @@ IF EXIST AUTORUN.BAT CALL AUTORUN.BAT
     }
     restart() {
         this.clearEJSResetTimer();
-        this.functions.restart();
+        this.EJS.withCoreAudioCapture(() => this.functions.restart());
     }
     getState() {
-        return this.Module.EmulatorJSGetState();
+        const info = this.Module.UTF8ToString(this.functions.saveStateInfo()).split("|");
```
ethanaobrien, line 211: Isn't this the same code that EmulatorJSGetState is lol - if there's a speed issue here that I'm probably missing probably a better idea to upstream it to the RetroArch script then here

https://github.com/EmulatorJS/RetroArch/blob/170ffa717026f4cc1c17a6096d09a245e4f9b448/emscripten/emulatorjs.js#L3

data/src/emulator.js
```diff
@@ -1159,13 +1165,49 @@ class EmulatorJS {
             this.startGameError(this.localization("Failed to start game"));
         });
     }
+    withCoreAudioCapture(callback) {
```
ethanaobrien, line 1168: Did you attempt building RetroArch with set_volume exposed btw? Iirc this is how we had volume control before OpenAL. I'm not sure how I feel about proxying the audio context

ethanaobrien (changes requested):
Would you be able to add support to use openal if available?

ethanaobrien:
Sorry about conflicts

PR #1267, opened 2026-09-16
Set this.debug in EJS_Download

TRusselo:
EJS_Download's constructor never assigned `this.debug`, so its three `if (this.debug) console.log("Using cached version of", url)` statements could not run. A cache hit was indistinguishable from a miss without opening the Network panel.

The EmulatorJS instance is already passed to the constructor and carries the flag.

Verified in a browser: the line appears on a cache hit after the change, and never before it.


RobLoach:
Thanks for pushing things forward here. Had been worried the project was abandoned.

TRusselo:
> Thanks for pushing things forward here. Had been worried the project was abandoned.

https://github.com/TRusselo/scummvm-wasm

Fyi

ethanaobrien (approved):

ethanaobrien:
> Thanks for pushing things forward here. Had been worried the project was abandoned.

I've not been able to be as active as I like because of school and work

RobLoach:
Don't burn yourself out <3

PR #1269, opened 2026-09-16
Decode the core report so its options are honoured

TRusselo:
`downloadFile()` resolves the core report to `{ data, headers }`, where `data` is an `EJS_CacheItem` holding the raw bytes in `files[0].bytes`. The handler tests `typeof rep.data === "string"`, which is false, so it takes `rep = rep.data` and `rep` ends up as the cache item itself — `key`, `files`, `added`, `url`, and no `buildStart` or `options`. The JSON is never decoded.

Two consequences:

- `options.defaultWebGL2` is never read, so `webgl2Enabled` falls to `false` and **every core is fetched under its `-legacy` filename** regardless of what its report says. The regular filename is never requested by anyone.
- `Could not fetch core report JSON! Core caching will be disabled!` fires on every launch even when the report was fetched and is valid.

This decodes the bytes and parses them.

Tested against a build of this branch with no other changes:

| core | report | before | after |
|---|---|---|---|
| a core setting `defaultWebGL2: true` | has options | `-legacy` | regular filename |
| `fceumm` | `options: {}` | `-legacy` | `-legacy`, unchanged, still plays |

So only cores that actually declare the option change behaviour.


ethanaobrien (approved):
Nice catch, thanks for the fix

PR #1270, opened 2026-09-16
Show errors in red and stop styling every message as one

TRusselo:
`.ejs_message` is styled `color: red`, so every message is an error to look at -- "SAVED STATE TO SLOT 1" reads as a failure.

`displayMessage()` takes an optional severity that colours the element, and the save-state and netplay call sites pass it. A caller that passes nothing gets what it always got, minus the red.

- base: white
- success: green (saved/loaded state)
- error: red (failed to save state, netplay connection/join/room errors)

The class is cleared when the message expires and replaced on each new message, so a later message never inherits the previous colour.

No behaviour change beyond colour, and independent of #1267, #1268 and #1269.


ethanaobrien:
The only concern I have on this PR is readability - could you add a faded backdrop to the text when it's on screen?

ethanaobrien:
Sorry about conflicts

TRusselo:
Added in acf3548a6 — `.ejs_message:not(:empty)` gets `rgba(0, 0, 0, 0.55)` with a 4px radius, shrink-wrapped to the text since the element is absolutely positioned with no width. `:empty` is hidden so an empty box never shows over the video.

Sorry for the slow reply — that went in a few hours after your comment and I never said so.

The `check-newlines` failure is on `data/localization/zh.json` and `data/src/utils.js`, neither of which this PR touches. #1272 fixes both and clears it.

PR #1271, opened 2026-09-16
Show decompression progress instead of a frozen download percentage

TRusselo:
The loading text holds the last download value -- usually "100%" -- for the whole unpack, which on a large archive is most of the wait.

Everything needed for this is already in the tree:

- `cache.js:236` and `:242` call `onProgress("decompressing", ...)`
- `cache.js:100`'s docstring says the callback "returns status(downloading or decompressing)"
- `Decompress Game Core` / `Data` / `BIOS` / `Parent` / `Patch` are translated in 22 localization files

but `emulator.js`'s progress wrapper drops every decompressing tick with `if (status === "downloading")`, so none of those strings can ever be displayed.

The wrapper now passes the phase through, and the two call sites with both a progress readout and an existing translated string are relabelled. `Download Game State` is left alone -- a `.state` file is not an archive, and there is no `Decompress Game State` string.

One detail worth flagging: the extractor reports `percentage` with `total` and `loaded` both zero, so the old `total ? ... : ...` ternary would have shown `0.00MB` for the entire decompression. `percentage` is now honoured on its own.


ethanaobrien (approved):
Nice catch, thanks for the fix

PR #1272, opened 2026-09-20
Add the missing trailing newlines

TRusselo:
`check-newlines` scans the whole tree rather than the diff, so `data/localization/zh.json` and `data/src/utils.js` fail it on every pull request, regardless of what that pull request changes. The run on #1270 is an example: it touches neither file.

Both files gain a single trailing newline and nothing else. `zh.json` still parses as JSON and `utils.js` still parses as a module.

AI assistance: written with Claude, reviewed and tested by me before opening.


TRusselo:
Nudge on this one — it fixes exactly the two files `check-newlines` currently fails on (`data/localization/zh.json`, `data/src/utils.js`), so merging it also clears that check on #1270 and #1273. Green here, and rebased on current main.

TRusselo (commented):
looks good

PR #1273, opened 2026-09-20
Version the core URL with buildStart

TRusselo:
`downloadGameCore()` caches the core keyed on `buildStart` and correctly decides a stored copy is stale after a rebuild. It then refetches:

```js
const corePath = "cores/" + filename;
```

a stable URL, so the HTTP cache returns the copy it already had and the stale core is reinstated. The report immediately above is cache-busted hourly for exactly this reason; the core it describes is not.

This puts the same version in the core's query, at the line that already has it in scope:

```js
const corePath = "cores/" + filename + "?v=" + encodeURIComponent(rep.buildStart);
```

A rebuilt core becomes a different request rather than a cache hit on an old one, and cores can then be cached long-term the way content-hashed assets are.

**No new eviction concern.** `EJS_DownloadCache` looks up by URL and stores under a content hash, with LRU eviction under `maxSizeMB` (4096 by default), so a superseded core is reclaimed like any other entry. The same build still hits, since the URL is unchanged.

**Tested** in a deployment that rebuilds its core frequently: before this, a replaced core kept being served from cache until site data was cleared by hand; after it, the new core is fetched on the next launch.

Worth noting separately: with the core versioned, the report's hourly bust becomes the bound on how quickly a new core is noticed rather than a correctness mechanism. I have left it alone as out of scope for this change.

AI assistance: written with Claude, reviewed and tested by me before opening.


TRusselo:
The `check-newlines` failure here is on `data/localization/zh.json` and `data/src/utils.js`, which this PR does not touch — #1272 fixes both.

